PR #8, opened 2016-09-25
only display '[Read More]' link if post is truncated

tomhoover:
The beautifulhugo theme includes a '[Read More]' link, even on short posts that are not truncated. For example, both 'Soccer' and 'Pirates arrrr' display the '[Read More]' link; however, 'Pirates arrrr' is displayed in its entirety and therefore doesn't need to display the link: 

screenshot 2016-09-25 08 41 34

This pull request modifies the theme to exclude the '[Read More]' link when the post is short enough to be displayed in its entirely (i.e. not truncated):

screenshot 2016-09-25 08 51 33 2

PR #9, opened 2017-02-06
Allow to see the warning when the layout is unspecified

1138-4EB:
'NO MATCHING PARTIAL' is hidden under the navbar, unless <header.html…> is printed before. It seems to be no problem moving its insertion frome <post.html> and <page.html> to <single.html>, so that it is always shown.


1138-4EB:
I merged the other pull requests.
On top of that, when deployed some links where broken in the head, due to a missing slash.
Moreover, when deploying to gh-pages I've realized that https has to be used to load font-awesome a gfonts.

halogenica:
I just wanted to say thank you so much for the comprehensive changes!

1138-4EB:
You are welcome. The remaining PR #8 can also be closed, since it is already merged:
Merge branch 'right-justify-readmore' of https://github.com/tomhoover…  fb026dc

PR #12, opened 2017-02-18
only display '[Read More]' link if post is truncated

tomhoover:
The beautifulhugo theme includes a '[Read More]' link, even on short posts that are not truncated. For example, both 'Code Sample' and 'First Post!' display the '[Read More]' link; however, 'First Post!' is displayed in its entirety and therefore doesn't need to display the link: 

screenshot 2017-02-18 12 34 38

This pull request modifies the theme to exclude the '[Read More]' link when the post is short enough to be displayed in its entirely (i.e. not truncated):

screenshot 2017-02-18 12 46 49

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -48,8 +48,12 @@
               </p>
             
               <div class="post-entry">
+              {{ if .Truncated }}
                 {{ .Summary }}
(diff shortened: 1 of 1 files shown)


halogenica:
Good change, thank you!

PR #18, opened 2017-02-26
Use Bootstrap ListGroup for terms overview

haidelber:


halogenica:
Looks wonderful, thank you!

PR #27, opened 2017-03-08
Fix bigimg issues

liwenyip:
- Fixed issue where bigimg only works if at least 2 bigimgs are defined (changed `if isset .Params.bigimg 1` to `if isset .Params.bigimg 0`)
- Fixed issue where title and subtitle disappear when bigimg is enabled and screen width is
less than 365px

This time I've created the pull request from a separate branch of liwenyip/beautifulhugo, which I believe means that if I push more updates to my master it won't pile those additional changes on this pull request?


liwenyip:
I only made changes to index.html and wp_header.html - please ignore the changes to the to the other files (not sure what's going on - I thought I'd merged halogenica:master back into liwenyip:master before creating this branch)

liwenyip:
Ignore previous comment - I managed to get those other files back in sync with halogenica:master. (Still don't know why my GitHub desktop thought they were already in sync - had to sync those files manually i.e. by downloading the latest versions from the GitHub website. I'm still new to Git...)

halogenica:
No worries, I'm still learning git too. Unfortunately I don't know why you're having issues with your sync's though, wish I could help.

Thanks for fixing the off-by-1 error; I merged this in here 917f5fef57fefd35cdde8900fa2f39be30fa19b3

Regarding the missing title and subtitle, I don't think I've been able to reproduce that problem. And looking at your fix, it seems that you just unconditionally program a second title, subtitle, or header partial (so if a bigimg was defined you would define the title/subtitle in that if block and then again outside that if block). I'm not sure if that's what you meant to do, or why that's an appropriate fix since I wouldn't expect it to be set twice? I'll await your response before closing this pull request.

liwenyip:
To reproduce, view www.liwen.id.au on a phone with screen width less than 365px or using device mode in Chrome:

# Incorrect behaviour:
image

# Correct behaviour:
image

The `div.intro-header` is indeed set twice if there is a bigimg - once as `<div class="intro-header big-img">` and once as `<div class="intro-header no-img">` - this is the way it's done in beautiful-jekyll [header.html](https://github.com/daattali/beautiful-jekyll/blob/ea6c0ad8342866916f6d545a8a74b0d3d4e59f01/_includes/header.html)

If you search main.css for `365px` you'll see the code that decides whether to show `div.intro-header.no-img` or `div.intro-header.big-img` depending on screen width:

```css
.header-section.has-img .intro-header.no-img {
  display: none;
}
@media only screen and (max-width: 365px) {
  .header-section.has-img .intro-header.no-img {
    display: block;
  }
  .intro-header.big-img {
    width: 100%;
    height: 220px;
  }
  .intro-header.big-img .page-heading,
  .intro-header.big-img .post-heading {
    display: none;
  }
  .header-section.has-img .big-img {
    margin-bottom: 0;
  }  
} 
```
There is probably a more efficient solution, but this way works.

halogenica:
Thank you for the detailed explanation! I really appreciate it. I've checked in the last of that code here: f1c6df1122f5c40972e8d22e6ee8befdfe1a8908

PR #30, opened 2017-03-11
Support static content on front page (above posts)

liwenyip:
#Demo:

image


liwenyip:
@halogenica I just realised that this shows `_index.md` on every page of the homepage (i.e. `/` and `/page/2/` etc).

Do you think by default `_index.md` should only be shown on page 1 of the homepage i.e. only on `/`?

Or should we make this an option?

PR #35, opened 2017-03-21
Photoswipe

halogenica:
Initial support for PhotoSwipe


liwenyip:
Would you believe I just published this: https://github.com/liwenyip/hugo-easy-gallery/
Demo at https://www.liwen.id.au/heg/

halogenica:
Awesome! This looks great, I'll work on integrating that soon and updating the links to your new pages.

halogenica:
Integrated and looking great, thanks for the awesome set of effects! 19b620f632eb075c6064acb3c0c475e3c5f18f50

You should tell spf13 that you wrote hugo-easy-gallery; he may link to it here http://gohugo.io/tools/ (in the "Other" section at the bottom). I'm sure many people would like to use it. 

I was wondering, is there a good (and backwards compatible way) to define different text on hover versus the captions you'd see when you click on the image. For example when I have a long description, it doesn't look too good to have that description appear on hover. Any suggestions? I was thinking the "alt" property might be a good standard, and possibly prefer "alt" for the hover text over the "caption" text if alt is defined (otherwise use the caption text). Not sure if that would be considered unexpected behavior though.

liwenyip:
> I was wondering, is there a good (and backwards compatible way) to define different text on hover versus the captions you'd see when you click on the image. For example when I have a long description, it doesn't look too good to have that description appear on hover. Any suggestions? I was thinking the "alt" property might be a good standard, and possibly prefer "alt" for the hover text over the "caption" text if alt is defined (otherwise use the caption text). Not sure if that would be considered unexpected behavior though.

That's a good idea, and I think using "alt" will work, just need to think it through when I have some more time.

Can you please submit an issue at https://github.com/liwenyip/hugo-easy-gallery/issues to make sure I don't forget?

Thanks in advance

liwenyip:
>You should tell spf13 that you wrote hugo-easy-gallery; he may link to it here http://gohugo.io/tools/ (in the "Other" section at the bottom). I'm sure many people would like to use it.

Will do, thanks 👍

PR #38, opened 2017-03-25
Use .Summary instead of .Content for og:description and twitter:description

liwenyip:
`.Summary` is much simpler than `.Content | plainify | htmlEscape | truncate 200` and it also automatically cuts off the content after `<!--more-->`


halogenica:
Excellent, thank you!

PR #39, opened 2017-03-25
Added support for Google Custom Search Engine

liwenyip:
Added support for Google Custom Search Engine using instructions at  https://www.cambiaresearch.com/articles/84/how-to-integrate-a-google-custom-search-popup-in-a-bootstrap-navbar

Live demo at https://www.liwen.id.au/

NB to make sure everything behaves properly, it's very important to set Layout to "Full Width":

image

And set theme to "Classic":

image

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -16,6 +16,7 @@ pygmentCodeFences = true
   commit = false
   rss = true
   comments = true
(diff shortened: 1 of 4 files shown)


halogenica:
Hey just FYI, there was a load error with the cse.js if the gcse wasn't configured in config.toml. I made a quick fix here: de52bb63ed7e5fba539ae218061d2740db40d160

PR #40, opened 2017-03-26
Add default "posted on" text in a new partial template

rayjolt:
If sites don't have `.Site.Params.dateFormat` set, then Hugo aborts
when trying to render `{{ .Date.Format .Site.Params.dateFormat }}`,
leading to the index page missing post previews and the footer,
and post pages having all the content styled with the "post-meta"
class.

To fix this, specify a default value, and put it in a new partial
template so that there is only one place we have to change it if
we need to.

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -39,7 +39,7 @@
                 </a>
       
                 <p class="post-meta">
-                  Posted on {{ .Date.Format .Site.Params.dateFormat }}
+                  {{ partial "posted-on-date.html" . }}
                 </p>
                 <div class="post-entry">
                   {{ if .Truncated }}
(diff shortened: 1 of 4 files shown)


halogenica:
Thanks for the change! I added a default date format here 4d866ddfd7f817684eda39bb7a5f320b1a7930c5. I did not add the partial, because the single place for a user to define a default is already the `config.toml` file, and I think that single location is simpler.

rayjolt:
Thanks. :) No worries about the partial. I was thinking of having one place for the default date to be defined, not the user-specified date, but I guess three places isn't so bad. And if we decide to internationalise the theme (as I have been thinking of doing) then the default date would end up being in the i18n file anyway.

PR #43, opened 2017-03-28
Split the social icon data out from the footer template

rayjolt:
Add a "social.toml" data file to hold the data for all the social
icons in the footer template. This reduces repetition in the
template code and makes it easier to reuse the data in other
parts of the theme, should that be desired in the future.

It also makes it easier for end users to add new icons by just
overriding the data file, instead of having to override the whole
footer template.

### DIFF
--- a/data/social.toml
+++ b/data/social.toml
@@ -1,0 +1,96 @@
+[[social_icons]]
+id = "email"
+url = "mailto:%s"
+title = "Email me"
+icon = "fa-envelope"
+
+[[social_icons]]
+id = "facebook"
+url = "https://www.facebook.com/%s"
+title = "Facebook"
+icon = "fa-facebook"
+
+[[social_icons]]
+id = "googleplus"
+url = "https://www.plus.google.com/%s"
+title = "Google+"
+icon = "fa-google-plus"
+
+[[social_icons]]
+id = "github"
+url = "https://github.com/%s"
+title = "GitHub"
+icon = "fa-github"
+
+[[social_icons]]
+id = "twitter"
+url = "https://twitter.com/%s"
+title = "Twitter"
+icon = "fa-twitter"
+
+[[social_icons]]
+id = "reddit"
+url = "https://reddit.com/u/%s"
+title = "Reddit"
+icon = "fa-reddit-alien"
+
+[[social_icons]]
+id = "linkedin"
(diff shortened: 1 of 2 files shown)


halogenica:
Thanks! I'll integrate this. Had a couple questions: 
- Is the `/data/` directory with .toml data files a standard thing? I've never seen this before.
- What do the `-` dashes mean in the lines `{{- if isset $.Site.Author .id }}` `{{- end -}}`?

rayjolt:
To answer your questions:
- Data files can be TOML, YAML, or JSON. They're documented [in the Hugo docs under "extras"](https://gohugo.io/extras/datafiles/). Also, I when I looked at that page again just now it suggests using `mytheme/data/mytheme/myfile.yaml` instead of `mytheme/data/myfile.yaml`, so we should probably do that.
- The `-` dashes trim whitespace. `{{-` trims the whitespace immediately before the braces, and `-}}` trims the whitespace immediately after the braces. (That's in the docs [here](https://gohugo.io/templates/go-templates/#whitespace).) I added them so that the rendered HTML would look a bit neater if you don't specify all the icons. I left the final braces of `{{- if isset $.Site.Author .id }}` without a dash so that we still get one newline between a closing `</li>` and the next opening `<li>`.

halogenica:
Thanks for the info! I read the hugo docs, and it actually sounds like you would specifically want to add the additional "mytheme" directory to the path of the data file specifically if the data was never meant to be overridden (since a user of the theme would define their own data files in the root of the data folder). Since in your example above you explained that end users may want to add their own social networks and I think they might even want to customize their own icons, you would want your data in a place the end user is easily able to override. Which would mean leaving the data file exactly where you have it.

I think this looks good, I'll merge it now. Thanks again!

rayjolt:
I think what the documentation was getting at is that a user might have a data file called "social.toml" already, and be using it for a completely different purpose. If they do, then that file will override the data file in the theme. That would mean that footer.html wouldn't be able to find any of the data for the icons, and it would cause a bunch of errors. If users specifically want to override the Beautiful Hugo data, then they can still do that by adding their own `data/beautifulhugo/social.toml` file.

halogenica:
Ok, fixed: cfad05b6a02970de49ef8cdab71d64ed42173c9a

PR #45, opened 2017-03-28
Update the footer template with the new data file location

rayjolt:
The social icons data was moved in cfad05b, but we need to update
the footer template with the new location as well to be able to
access the data.


halogenica:
Thanks :) mine was working locally just because I had tested overriding the data in my demo website.

PR #63, opened 2017-05-10
Add Chinese language support

lzjluzijie:


lzjluzijie:
Sorry, i did not test the date format. It should be "January 2, 2006".

ginuerzh:
I think for Chinese users the date format such as `2006-01-02 15:04` may be more natural.

PR #64, opened 2017-05-14
Avoid using css import

esaezgil:
Avoid using css `@import` in main.css file 
Some context: https://varvy.com/pagespeed/avoid-css-import.html


esaezgil:
Thanks you are totally right, I submitted the MR too soon, that's also why I closed it asap. After your comment I found that the link element is already in `head.html` and this is duplicated:
https://github.com/halogenica/beautifulhugo/blob/master/layouts/partials/head.html#L47

I checked with the result of removing it and everything works as expected :)

PR #80, opened 2017-07-19
added authors website to site parameters. link to authors website can…

ROYALBEFF:
… be included in the footer now.

We are a students club and use this theme for a website of one of our projects. We set the author-name parameter to the name of our club and want to add the link of the clubs website ot the name in the footer. So we added an additional parameter author-website and changed a few lines in the footer.html. The author-website paramter is optional and can be ignored if not needed.

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -31,6 +31,7 @@ pygmentCodeFences = true
 
 [Author]
   name = "Some Person"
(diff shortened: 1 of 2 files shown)


halogenica:
Thank you for the PR! I'm very glad to hear you were able to use the theme for your projects :)

PR #83, opened 2017-08-02
generate tag links with trailing slash

cite:
Since all post URIs are generated with a trailing slash, I thought
it would be a good idea to do the same for tag links. This can
save a few I/O operations here and there (or, depending on your
webhoster, even get rid of redirects).


cite:
I just realized I did this from master - my bad.

PR #85, opened 2017-08-05
some URL related fixes, mainly tags

cite:
* generate all links to tags with trailing slash (saves I/O on webserver)
* generate all links to tags with complete URL (for when website doesn't live in domain root)
* include load-photoswipe.js from absolute URL (for when website doesn't live in domain root)


halogenica:
Thank you very much for the additions! I appreciate it.

PR #86, opened 2017-08-06
include complete URL in tags

### LINKED ISSUE #71: Tag URL
I create a change for baseurl, from `domain.com` into `domain.com/dir`, i try to change every link and work properly except the tags.

The main tag (i don't know what i called) that appears in the main page is work. but the Tag from list (in the tag menu) is not working. till using `domain.com/tag/thetag` that should be `domain.com/dir/tag/thetag`.

It maybe my bad who don't know how to configure it, but maybe you could give some help.

Thanks.

cite:
I want to resolve #71 with this - links to tags should now also
work when the site is not living in a domain's root.

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -51,7 +51,7 @@
               {{ if .Params.tags }}
                 <span class="post-meta">
                   {{ range .Params.tags }}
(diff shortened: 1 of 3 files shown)


halogenica:
Thank you very much!

PR #87, opened 2017-08-09
corrected variable name for authors website in footer layout

ROYALBEFF:
Found a mistake in the footer layout file. The if that checks for the authors website used a wrong variable name (.Site.Author.web instead of .Site.Author.website).


halogenica:
Thank you!

PR #88, opened 2017-08-24
Create Spanish Translation.

ccamara:
Added yaml with Spanish translation. As I have just discovered Hugo and just started to use it and test it, I don't know if something else is needed, though, for a proper Spanish translation.

### DIFF
--- a/i18n/es-ES.yaml
+++ b/i18n/es-ES.yaml
@@ -1,0 +1,39 @@
+# Content
+- id: dateFormat
+  translation: "2 de enero, 2006"
+- id: postedOnDate
+  translation: "Publicado el {{ .Count }}"
+- id: translationsLabel
+  translation: "Otros idiomas: "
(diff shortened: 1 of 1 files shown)


i18n/es-ES.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+# Navigation
+- id: toggleNavigation
+  translation: "Conmuta navegación"
+- id: languageSwitcherLabel
+  translation: "Idioma"
+- id: gcseLabelShort
+  translation: "Buscar"
+- id: gcseLabelLong
+  translation: "Search {{ .Site.Title }}"
```
esaezgil: this should be  `"Buscar {{ .Site.Title }}"`

ccamara:
Thanks. Missing string was added.

halogenica:
Thank you very much!

PR #89, opened 2017-09-04
generate page links with trailing slashes

cite:
This will generate links to pages with trailing slashes, e.g. `/page/2/` instead of `/page/2`. Again, this saves I/O ops :-)


halogenica:
Thank you very much!

PR #94, opened 2017-11-08
Add chroma highlighting support (Hugo now use it)

badele:


henryiii:
I was just about to make this change and make a PR! Thanks! 👍

henryiii:
I've made a few improvements, such as a note in the README on activating Chroma support in the theme (it's default in Hugo!) and theming. I removed the highlight JS (not just the CSS) when Chroma is activated, and am using the same name for the css file as in the docs. The only comment I might suggest is a better default; the monokai theme is ugly (IMO) and doesn't match the normal default for this theme. I'm locally using manni for now.

For the moment, @badele , I'll make the PR to your branch.

badele:
Thank @henryiii for you PR :)

halogenica:
Thank you both very much! I also wanted to add this, I appreciate your PR :)

PR #98, opened 2017-11-12
Add background-position to bigimg's configurable per image.

fuerst:
You may add a "background-position" value to a [[Params.bigimg]] entry if a big image needs positioning different from the default.


halogenica:
Thank you very much!

PR #100, opened 2017-11-18
More post information

badele:
Add more post information, if staticman used, it can also show number posted comments 

**note**:  you can see result in [my blog website](http://blog.jesuislibre.org)
**note bis**:  the preview pull request must be validated (merged from my previous staticman branch)


halogenica:
Thank you very much!

Issue #81, opened 2017-07-19
head_custom and footer_custom not uniform


ROYALBEFF:
The comments in the head_custom.html and footer_custom.html say to add the corresponding file to the `/layouts/` directory in order to add some extras to the head/footer. That's right for the head_custom.html. I tried the same for the footer_custom.html, but it didn't work. I figured out that moving the file to `/layouts/partials` fixes it. Both comments say the same, but it just works for one of the files. It would be nice if both files should be either in `/layouts/` or in `/layouts/partials/`. I would have fixed it myself, but I think in the end you should decide where to put the files.

jkinred:
I found that the path didn't work for either, raised PR to correct path.

halogenica:
Thanks @ROYALBEFF for letting me know, and thank you @jkinred for the fix! aa21f82dff26bd63326220093fc81970fe86c242

PR #109, opened 2017-11-26
Update doco in head_custom.html and footer_custom.html to correct path

jkinred:
As per #81. I verified with `strace` that the paths referred to are never read from and only `layouts/partials/` is search by Hugo.


halogenica:
Thank you very much!

Issue #105, opened 2017-11-22
Translation of "words"


podiki:
Right now several (all?) languages in the i18n yaml files use the word "mots" for the translation of "words." Much as I love French, this should be updated.

badele:
It's my fault, I have created a new feature (post information) and i forgot to translate that word, I made a PR

PR #112, opened 2017-12-07
Add modified time, fix translations, make reading time optional

henryiii:
The recent changes to the master branch have introduced a few issues, this should fix those.

The fairly newly added reading time is now optional via `readingTime=true` parameter, and is correctly translated (English version does not have French in it now, and other translations added via Google Translate).

The modification date has also been added, and is only shown if different from the creation date (so it will not clutter a site unless the author is using modification times)

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -17,6 +17,7 @@ pygmentCodeFences = true
   commit = false
   rss = true
   comments = true
+  readingTime = true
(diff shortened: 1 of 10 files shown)


henryiii:
Also fixes #105

dadosch:
But the German translation "wörter" should be capitalised like that: "Wörter"

henryiii:
Fixed, thanks!

halogenica:
Thank you for these changes! Sorry I missed them in my previous commit, your changes were better :) Should have them merged here: e0a5a06ce19a6260a09ce47fed4946016d0c83e9

Issue #108, opened 2017-11-24
[Question]: Hiding next | previous post buttons on static pages


arwhyte:
The Demo Beautiful Hugo site includes a static "About" page that hides the next | previous post-> button that is associated with blog entries. [1]  I'd like to hide the "post" button on my static pages as well but when I mimic the "About" page with my own "About" page the post button is displayed on it.   How does one  hide/exclude the next | previous post-> button on static pages?

My initial content directory
```
content
  page
    about.md
  post
    post1.md
   . . .
```

[1] https://themes.gohugo.io/theme/beautifulhugo/page/about/

johanbrandhorst:
Also running into this issue :(

henryiii:
If this solution is better, I can change my PR to do this instead.

halogenica:
Thank you @arwhyte and @johanbrandhorst for pointing this out, and big thanks to @henryiii for fixing it! Merged: 9eea743863beb9cfb7cbe6eee4d38572a00e5471

PR #113, opened 2017-12-07
Remove next/previous for pages

henryiii:
Pages have next/previous buttons added just like they were posts. This drops that if the type is page. This also removed the site-level setting for comments if the type is page (an individual page can override this by setting `comment=true` on the page if the author really wants page comments).

You can manually set:
```
type: page
```

or it will be taken from the section (folder) name ([from the Hugo docs](https://gohugo.io/variables/page/)).

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -6,7 +6,7 @@
         {{ .Content }}
       </article>
 
-      {{ if ne .Params.type "page" }}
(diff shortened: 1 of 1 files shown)


henryiii:
I just saw that this would fix issue #108 as well.

halogenica:
Thank you very much!

PR #115, opened 2017-12-09
Better spacing around paragraphs

henryiii:
This adjusts a few spacings so that the large space between paragraphs only shows up between subsequent paragraphs, not before all paragraphs. Makes things like code boxes much nicer.

(Also added a bit more space above the next/previous buttons)

You can see an example of the new spacing and other changes here: https://iscinumpy.gitlab.io

(The only change I've not made a PR for is the archive listing; it's not quite ideal and general, though if someone is interested in working on it, my source is available.)

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -16,7 +16,10 @@ body {
 }
 p {
   line-height: 1.5;
-  margin: 30px 0;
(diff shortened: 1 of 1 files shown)


halogenica:
Thank you very much!

PR #116, opened 2017-12-09
Expanding taxinomies

henryiii:
This add bootstrap expanding taxinomies. When you click a tag or category, the post list appears inline. Please look at it and decide if you like it before merging, as the style is a little bit different. Feel free to update or suggest if you like something better!

https://iscinumpy.gitlab.io/tags/ is an example.


jl5000:
I like it, works well!

nogu-net:
I like it, but doesn't work using CJK characters in tag.
Problem entry is here.

CJK-test.md
---
title: "てすと"
date: 2018-01-24
draft: false
tags: [てすと]

---
てすと

nogu-net:
It works well modified line no 11.
<pre>
<a class="collapsed" role="button" data-toggle="collapse" data-target="#collapse{{ $value.Name }}" data-parent="#accordion" href="#" aria-expanded="false" aria-controls="collapse{{ $value.Name }}">
</pre>

henryiii:
@nogu-net, thanks! I've updated my website branch to use the fix, as well.

nogu-net:
@henryiii, Thanks!

halogenica:
I like this too :) I'm thinking of a way the UI could have both behaviors; like clicking on the tag name loads a list page, but clicking on a `+` expands them.

halogenica:
I like it enough with "View All" that I'll just merge this now. And maybe I'll add a config option to enable/disable. Very cool stuff, thank you very much!

PR #124, opened 2018-01-11
update PhotoSwipe to v4.1.2

tmo-github:
From the release notes:
"Fix: iOS 10.3 not updating layout after orientaton change in some cases."

I have only updated these 4 files, which seem to be the only ones referencing PhotoSwipe in your theme.


halogenica:
Thank you very much!

PR #127, opened 2018-01-19
Fix missing slash in tags URL is also in layouts/_default/list.html.

nogu-net:
Fix missing slash in tags URL is also in layouts/_default/list.html.


halogenica:
Thank you very much!

PR #130, opened 2018-02-09
Add `post_meta.html` reference to `_default/list.html`

joshuasheppard:
It looked like the `layouts/_default/list.html` was missing a reference to `post_meta.html` to match what was already in `index.html` and `layouts/partials/header.html`. Otherwise list pages won't use the same formatting as many others.

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -37,7 +37,7 @@
               </a>
 
               <p class="post-meta">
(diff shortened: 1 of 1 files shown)


joshuasheppard:
Please let me know if there is more I should do to make this an acceptable PR. I'm new to contributing here and didn't see a specific note about how you'd like to see PR's handled.

halogenica:
Excellent, thank you very much!

PR #133, opened 2018-02-12
add social network share links

jms:


johanbrandhorst:
This is great! @halogenica any chance we could have this merged?

hukacode:
There is one wrong type, Pintrest should be Pinterest.

dancwilliams:
Running this on my site ([`https://networkhobo.com`](https://networkhobo.com)) and it works great!  Thanks for the contribution!

johanbrandhorst:
On second thought, I actually think this could do with some more configuration options. I made a some custom sharing buttons with the use of https://sharingbuttons.io/ and it was extremely easy to integrate, so I think it should be possible to create a partial that allows customization beyond these.

halogenica:
Wonderful addition, thank you very much!

PR #135, opened 2018-02-12
Fix error when there is no comment in staticman -- Fixes #134

### LINKED ISSUE #134: Get an error when there are no comments in staticman
We get an error where there are no comments in staticman

` executing "main" at <partial "staticman-c...>: error calling partial: template: theme/partials/staticman-comments.html:19:33: executing "theme/partials/staticman-comments.html" at <index $.Site.Data.co...>: error calling index: index of untyped nil
`

tutysara:
Fixes #134


halogenica:
Thank you very much!

tutysara:
yw, and thanks for the beautiful theme

PR #138, opened 2018-02-18
Allow title in _index.md

miketweaver:
This allows you to set a custom title in `_index.md` apart from the `Site.Title`

The purpose of this is to be able to customize the `<h1>` text on the index page without affecting the rest of the site where the `Site.Title` is used.

The purpose of using `isset .Params "title"` in the code instead of `if .Title` is that it allows you to set `title = ""` and have that `<h1>` title disappear without removing the text from `<title></title`


halogenica:
Thank you very much!

PR #143, opened 2018-02-21
HTTPS everywhere !

tmo-github:
A little cleanup to use secure URLs and require less browser redirections (links tested)


halogenica:
Thank you very much!

PR #147, opened 2018-03-11
Faster FontAwesome loading.

excitoon:
Kind regards,
Vladimir.


halogenica:
I've had a request to replace fontawesome with forkawesome (https://github.com/halogenica/beautifulhugo/issues/152); so I expect to move in that direction. Thank you very much for the PR!

PR #154, opened 2018-04-01
fix pagination to respect PaginatePath configuration option

NuLL3rr0r:
Please take a look at https://gohugo.io/content-management/urls/

Current pagination url ignores PaginatePath set by user in configuration file which is wrong:

My config.toml looks like this:

```
Paginate = 8
PaginatePath = "/blog/page"
```

Hugo generates my pages at "/blog/page" but Beautiful Hugo Next and Previous buttons always point to /page/2 /page/3, ... which obviously does not respect the user's choice and only works for the default /page value.

This commit fixes that, so whatever is the PaginatePath's value the pagination works.

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -49,12 +49,12 @@
           <ul class="pager main-pager">
             {{ if .Paginator.HasPrev }}
               <li class="previous">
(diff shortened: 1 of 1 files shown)


halogenica:
Good catch, thank you!

PR #169, opened 2018-05-07
Rename figure to beautifulfigure

### LINKED ISSUE #114: Figure shortcode issues
I had some problems with the figure shortcode, since it doesn't support all the features of the original figure shortcode (I was either trying to resize or center a figure). I ended up removing it. Would it be better to rename it to something like `beautifulfigure` so that the original figure shortcode is accessible?

henryiii:
This fixes #114 by renaming figure to beautifulfigure. That way, a figure is still usable without adding all the gallery type additions that beautifulfigure adds. For centered figures, diagrams, etc., this is much cleaner.

I picked `beautifulfigure`, but other ideas include `picture`, `gallery`, and `fancyfigure`. This is a fairly intrusive change, but there currently is no way for a normal user to restore the default "figure" environment without modifying beautifulhugo, and figure != gallery for some (most?) users.

### DIFF
--- a/exampleSite/content/post/2017-03-20-photoswipe-gallery-sample.md
+++ b/exampleSite/content/post/2017-03-20-photoswipe-gallery-sample.md
@@ -8,9 +8,9 @@ tags: ["example", "photoswipe"]
 Beautiful Hugo adds a few custom shortcodes created by [Li-Wen Yip](https://www.liwen.id.au/heg/) and [Gert-Jan van den Berg](https://github.com/GjjvdBurg/HugoPhotoSwipe) for making galleries with [PhotoSwipe](http://photoswipe.com) . 
 
 {{< gallery caption-effect="fade" >}}
-  {{< figure thumb="-thumb" link="/img/hexagon.jpg" >}}
-  {{< figure thumb="-thumb" link="/img/sphere.jpg" caption="Sphere" >}}
-  {{< figure thumb="-thumb" link="/img/triangle.jpg" caption="Triangle" alt="This is a long comment about a triangle" >}}
+  {{< beautifulfigure thumb="-thumb" link="/img/hexagon.jpg" >}}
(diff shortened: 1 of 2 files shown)


halogenica:
I definitely want to allow users to disambiguate between Hugo's figure shortcode and this theme's figure. Your change is probably the best way to do it, I'm just trying to think about breaking backwards comparability. I might just have to rip off the band aid :)

halogenica:
Just FYI I've been cleaning up my backlog of PRs and issues recently, and I'm waiting to see if there's other content-breaking changes that I can merge at the same time as this just to keep thrash low. I do plan to merge this :)

VincentTam:
I have browsed `gallery.html` and `beautifulfigure.html` at this Git commit.  They are *outdated* as this doesn't include liwenyip/hugo-easy-gallery#10, which fixes the problem of loading figures in project pages.

henryiii:
I've based this on current master.

VincentTam:
Please take a look at PR 297, which removes the leftmost forward slash in `/images/foo/bar.png` in the file concerned.  The forward slash can cause errors in case of nonempty subpaths in `$.Site.BaseURL`, e.g. GitHub project pages.

Here's a direct quote from [Hugo Themes repo's *Common Permalink Issues*](https://github.com/gohugoio/hugoThemes#common-permalink-issues).

> Make sure not to use a forward slash `/` in the beginning of a URL, because it will point to the host root and Hugo will not generate the correct URL for the demo's assets.

In addition, @<span></span>bep says that "`/subpath/images/gopher.png` isn't very portable" in https://github.com/gohugoio/hugo/issues/4733#issuecomment-390200427.  A better approach would be using Hugo's methods like [`absURL`](https://gohugo.io/functions/absurl/), `relLangURL`, etc.

henryiii:
This is now in, in a backward-compatible way.

PR #170, opened 2018-05-07
Updating bootstrap to version 4

henryiii:
Everything seems to look okay.


pad92:
hello,

bootstrap is loaded from bootstrapcdn

pad92:
upgrade is done on this fork : https://github.com/joncutrer/eleganthugo and mine :)

halogenica:
Indeed, bootstrap is loaded from the CDN and not from these local files. The local files are there because I was eventually hoping to add local fallback functionality, but I've had trouble getting that working. I think there's probably a fair bit of work to get the theme ported over to bootstrap 4, but if anyone is motivated to do it and sees a strong benefit to the move, I'd welcome a PR :) Thanks!

PR #174, opened 2018-05-19
fix missing slash for tags -> "view all"

cite:
This won't probably show in "hugo serve" - all other occurences of `$.Site.LanguagePrefix | absUrl` seem to add a trailing slash, and it broke for me with a baseURL of `https://example.com` instead of `https://example.com/`.


halogenica:
Excellent, thank you very much!

PR #175, opened 2018-05-26
TranslationBaseName replaced for Name on archetypes template

nonumeros:
See https://github.com/gohugoio/hugo/commit/863a812e07193541b42732b0e227f3d320433f01


halogenica:
Thank you!

PR #177, opened 2018-06-06
Remove double foreward slash in tag links

pjbakker:
baseurl should contain the trailing slash after the domain name.

By having a slash in these links, actual links contain double slashes.

### DIFF
--- a/layouts/_default/list.html
+++ b/layouts/_default/list.html
@@ -51,7 +51,7 @@
               {{ if .Params.tags }}
                 <div class="blog-tags">
                   {{ range .Params.tags }}
-                    <a href="{{ $.Site.LanguagePrefix | absURL }}/tags/{{ . | urlize }}/">{{ . }}</a>&nbsp;
+                    <a href="{{ $.Site.LanguagePrefix | absURL }}tags/{{ . | urlize }}/">{{ . }}</a>&nbsp;
                   {{ end }}
                 </div>
               {{ end }}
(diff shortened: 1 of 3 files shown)


halogenica:
Thank you for the change! Just to educate myself, where is it documented that this is the expected way to define the domain name? I've merged a few PRs that "fixed" this the other way, which added slashes. I would like to know where the proper way to set this is documented so I can refer to it later when I inevitably get PRs to add these slashes back :)

Also, I strongly suspect that there are other URLs that need the slash removed; I can hunt for them all and make that change.

Thanks!

pjbakker:
I think the newest version does not require this anymore. The documentation did contain reference to this, but it's not there anymore. I see loads of the URL related code in Hugo has changes for the 'component' functionality.

halogenica:
Thanks for checking! Does that mean this should no longer be needed?

soofaloofa:
I tested this code very recently on an S3 bucket and it does not work with double slashes. I don't know if there is a canonical way to do this though but from what I understand S3 websites will fail if the key for an object begins with //.

soofaloofa:
Although someone else had this issue and just removed the slash from their baseurl. I can do that as well.

halogenica:
I just made a commit to make sure the theme uses absURL instead of BaseURL (which was one of the issues from your link). As far as I can tell, there's no consistency among themes about whether the URL in the config file is specified with a trailing slash. I try to be internally consistent with this theme, and if you specify your site without a trailing slash, everything should work. If that doesn't work, or if you find some other indication this is incorrect, please let me know :) For now, I'll close this PR. Thanks!

Issue #171, opened 2018-05-09
Summary text optionally instead of truncated body text?


fubarhouse:
In attempting to display alternative content on the homepage, I've noticed there's little control over influencing what is shown on the homepage.

I'd like to request for an optional summary field on posts to be able to provide alternative text explicitly as needed on the homepage.

Otherwise I'm loving the theme. 👍

halogenica:
Thanks for the request! Addressed here: 65cdb2e82d2ce73b44c51d1d06a64db6a4a7d134

PR #179, opened 2018-06-13
Allow custom summaries

zevdg:
This is an "escape hatch" that can be used when neither automatic nor manual summary splitting is sufficient to create a good summary.  This is specifically useful for articles that begin with a disclaimer or anything else that shouldn't be included in the summary.  It's also useful for footnote links and other mid-text items that should be excluded from the summary.

See https://github.com/gohugoio/hugo/issues/1634 for more use cases.


charly3pins:
#171

halogenica:
Thank you very much!

PR #180, opened 2018-06-28
Make sure one can override socialShare on a page by page basis

pjbakker:
By using $.Param instead of .Site.Params, the socialShare icon settings
can be overridden on a page by page basis in the Front Matter


halogenica:
Great idea, thanks for the PR!

PR #181, opened 2018-06-28
Add a default Description for taxonomy pages

pjbakker:
Make use of the detection of the .Kind of page to have a default
description for SEO purposes


halogenica:
Thank you very much!

PR #201, opened 2018-09-30
Delayed Disqus comments

igorkulman:
This PR adds an option to load the Disqus comments on demand when the user clicks a button. It saves the users a ton of network requests when they do not care about comments (or there are none for a post).

More info in my blog post [Delaying Disqus comments to save (a ton of) requests](https://blog.kulman.sk/delaying-disqus-comments-to-save-requests/) where you can also see it in action.

Delayed Disqus comments


halogenica:
Thank you very much!

Issue #137, opened 2018-02-15
codeblock.css Breaks Mobile Responsiveness


chris-short:
This is a real edge case. But, I recently was made aware that my site wasn't responsive on mobile. Testing led me to articles with code blocks being the only ones breaking responsiveness. After checking my config.toml (using Chroma, seemed to be fine) I had to deep dive the code. After significant troubleshooting, I determined the codeblock.css file was to blame. Purging that file rendered the site immediately mobile friendly. Not sure if it's just my site with this issue or if others are experiencing it as well.

aammd:
I have the same problem! this  post is just fine http://thestudyofthehousehold.com/post/twitter-recommends-stats-books/
but this one, which contains multiple code blocks, doesn't
http://thestudyofthehousehold.com/post/a-tidy-game-of-life/

I use quite a lot of code blocks, so this is quite a frequent problem for me. any advice you can offer would be wonderful! 
and thanks for making Dean's theme available in Hugo

chris-short:
The way I "fixed" it was to add a blank codeblock.css file to my static/css directory. Not sure where it's getting called at this point and I just wanted it fixed ASAP.

dancwilliams:
Just wanted to add what I did.

I added some extra detail to my `codeblock.css` file to enable code block scrolling on both desktop and mobile devices.  I tested across multiple platforms and everything seems to look good.

Here is my `codeblock.css` file.  You can test it out at [this link](https://networkhobo.com/2017/12/30/hugo---staticman-nested-replies-and-e-mail-notifications/) on my site [`https://networkhobo.com`](https://networkhobo.com). 

```css
/* --- Code blocks --- */

.chroma .ln { 
  margin-right: 0.8em; 
  padding: 0 0.4em 0 0.4em; 
}
pre code.hljs {
  padding: 9.5px;
}

pre code {
  overflow-wrap: normal;
  white-space: pre;
}

@media only screen and (min-width: 767px) {
  pre code {
    overflow: auto;
    word-wrap: normal;
    white-space: pre;
  }
}
```

PR #204, opened 2018-10-12
Improved codeblock rendering.

azriel91:
Makes code blocks with table style render nicer:
code_diff

* no scroll bar for line number cell
* rounded corners
* no middle border
* long code line doesn't wrap
* scroll bar for long code lines

cc: #137


halogenica:
This is wonderful, thank you so much for the fix! This was bugging me for a long time :)

PR #206, opened 2018-10-19
Improved code block rendering when not using line numbers

igorkulman:
Based on #206 but for the case when line numbers are not used.


halogenica:
thank you very much!

Issue #210, opened 2018-10-26
Have bigimg use page bundles


soofaloofa:
A common use case for bigimg is to have a per-post header picture that is unique to each post. Right now bigimg only supports images that reference the root level URL. It would be great if we could use relative paths and page bundles for this to simplify image management.

For example, the following toml configuration would reference `picture.jpg` in the current page bundle (current folder).

```toml
+++
bigimg = [{src = "picture.jpg", desc = "my picture"}]
+++
```

PR #211, opened 2018-10-30
Allow relative paths for bigimg.

soofaloofa:
This PR first checks for a bigimg at a full root URL (i.e., /post/content/...). If an image does not exist at that location, use a relative URL to find the img in the current posts page bundle.

Fixes: https://github.com/halogenica/beautifulhugo/issues/210


halogenica:
Great addition, thank you very much!

PR #215, opened 2018-11-11
Add Traditional Chinese language support

Gingfrederik:


halogenica:
This is fantastic, thank you so much!

PR #216, opened 2018-11-11
Updated post_meta.html to format the Author properly in else clause

dancwilliams:
In post_meta.html the else statement was not rendering the pipe and user icon.  Corrected this formatting.


halogenica:
Thank you very much!

PR #228, opened 2018-12-09
add useKatex flag to gate on katex loading

### LINKED ISSUE #163: Add the option to not use/load KaTeX.js
Because it affect loading speed and 99% of blogs do not render math.

Or maybe in Hugo it is possible to inject this library only when needed - according to posts scanning result.

clux:
Fixes #163

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -25,6 +25,7 @@ pygmentsCodefencesGuessSyntax = true
   readingTime = true
   wordCount = true
   useHLJS = true
+  useKatex = true
   socialShare = true
   delayDisqus = true
   showRelatedPosts = true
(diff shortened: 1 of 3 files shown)


clux:
Note that this changes the default to false, which is perhaps undesireable, but it keeps the naming convention in config.toml consistent.

igorkulman:
You forgot about the CSS in header: https://github.com/halogenica/beautifulhugo/blob/master/layouts/partials/head.html#L67

VincentTam:
There's a more backward compatible approach to this problem.

| | Direct | Indirect |
| --- | --- | --- |
| Param (defaults to `false`) | `useKatex` | `disableKaTex` |
| load KaTeX | `if useKaTeX` | `if !disableKaTeX` |
| Existing math sites | **broken** | untouched |
| Non-math sites | untouched | add `disableKaTex = false` |

PR #230, opened 2018-12-09
[Feature] Allow to use customised Month names in posted date month

paba19:
This is a workaround for missing localization in month names. Plus it allows a more generic approach by giving the possibility to rename the months however you want beyond translation.
    Once this code is in place in you can just set the config.yaml file as follows:

    languages:
      en:
        languageName: English
        title: "My awesome blog"
        weight: 1
      it:
        languageName: Italiano
        title: "Il mio fantastico blog"
        params:
          Months:
            1: "Gennaio"
            2: "Febbraio"
            3: "Marzo"
            4: "Aprile"
            5: "Maggio"
            6: "Giugno"
            7: "Luglio"
            8: "Agosto"
            9: "Settembre"
            10: "Ottobre"
            11: "Novembre"
            12: "Dicembre"

### DIFF
--- a/layouts/partials/post_meta.html
+++ b/layouts/partials/post_meta.html
@@ -1,9 +1,20 @@
 <span class="post-meta">
-  {{ $lastmodstr := default (i18n "dateFormat") .Site.Params.dateformat | .Lastmod.Format }}
-  {{ $datestr := default (i18n "dateFormat") .Site.Params.dateformat | .Date.Format }}
-  <i class="fas fa-calendar"></i>&nbsp;{{ $datestr | i18n "postedOnDate"}}
-  {{ if ne $datestr $lastmodstr }}
-    &nbsp;{{ $lastmodstr | i18n "lastModified"  }}
+  {{ if $.Site.Params.Months }}
+    {{ $monthDate :=  index $.Site.Params.Months (printf "%d" .Date.Month) }}
+    {{ $monthMod :=  index $.Site.Params.Months (printf "%d" .Lastmod.Month) }}
+    {{ $lastmodstr := printf "%d %s %d" .Lastmod.Day $monthMod .Lastmod.Year }}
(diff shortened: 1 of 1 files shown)


paba19:
Hi Vincent, thanks for your help, but I think the commit linked doesn't help in translating the months names.
However, I understand your point that rather than adding the month names in the config.yaml they should go either in i18n or in data and not be burden for the end user. 
The base idea was to implement something like: https://gohugo.io/content-management/multilingual/#customize-dates but creating a data file, is it not even a bigger burden for the end user? :) 
Also, please note, that with my proposed workaround, the user is *not* forced to add the months names in the config.yaml file! If they don't, it simply will keep the current behaviour.
I am still new to hugo, so I will try to see if I can add the names in the i18n files, so that the theme can ship with them already

VincentTam:
The data files are already there, under the `i18n` folder.  As a contributor of this theme's Chinese and Japanese UI texts for Staticman, I remember well that the YAML file for each language has 76 lines, and we support fourteen locales, giving 76 × 14 = 1064 lines.  Inserting 12 × 2 × 14 = 336 lines into the YAML files and adding appropriate long/short format for each of them in `i18n` is now a bruden for contributors.

However, you can help by suggesting an Italian translation for a long date format, or verify if "1 dicembre, 2014" is the appropriate translation.  Hopefully, with the collaborative effort, we'll make it some day.

paba19:
Hi sorry your reply went under the radar during these holidays.
Ok, feel free to close this PR if you wish. 
A proper translation would be Lunedí, 1 dicembre 2014. However also, let me know how I can contribute to help with this theme :-)

halogenica:
I agree that this should be translated via the i18n folder. I'm happy to accept a PR that adds this feature with the appropriate translations hunted down; but this exact PR is probably not the way to solve it so I'll close it for now. Thank you very much for the PR and suggestion, and I'm happy to merge another in the future!

PR #234, opened 2018-12-16
[Documentation] Changed to config.toml

VincentTam:
To match `config.toml` in `exampleSite`, its better to document the site config file in TOML.


halogenica:
Thanks for the PR!

Issue #240, opened 2019-01-14
Official directory is "posts", theme looks for "post" - causes posts to be missing by default


ventz:
Hi,

Opening a new ticket for this (since it's a separate issue), and referencing the "resulting" problem: https://github.com/halogenica/beautifulhugo/issues/150

It seems that `posts` is the official directory structure in hugo
(see official tutorial, as of latest 0.53: https://gohugo.io/getting-started/quick-start/)

@halogenica Would it be possible to to standardize on "posts"?

ventz:
@halogenica Created PR (https://github.com/halogenica/beautifulhugo/pull/241) with solution per recommendation from hugo about themes handling "posts" and "post" directories.

ibnunowshad:
May I ask, if this issue has been closed?

stefano-garzarella:
I also hit this problem, is there any update?

henryiii:
mainSections = ["post","posts"] handles both

PR #241, opened 2019-01-14
Updated posts rendering to recommended method, using '.Site.RegularPages Type in .Site.Params.mainSections'

ventz:
Fix for issue: https://github.com/halogenica/beautifulhugo/issues/240

Solution per official hugo recommendation for "posts" (or "post") type.

### DIFF
--- a/layouts/_default/baseof.html
+++ b/layouts/_default/baseof.html
@@ -8,6 +8,7 @@
     {{ block "header" . }}{{ partial "header.html" . }}{{ end }}
     {{ block "main" . }}{{ end }}
     {{ partial "footer.html" . }}
+    {{ block "footer" . }}{{ end }}
   </body>
 </html>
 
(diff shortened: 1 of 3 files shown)


ventz:
I have forked off `beautifulhugo` here: https://github.com/ventz/beautifulhugo
and I have applied this change!

pad92 (approved):

sagikazarmark:
What about https://github.com/halogenica/beautifulhugo/blob/master/layouts/partials/header.html#L46 ?

What should be the default there?

ventz:
@sagikazarmark That's a good question. I think the `.Type` is a different built-in and still supported, but I am also not sure.

sagikazarmark:
I rather meant the hardcoded `"post"` type.

henryiii:
I believe all fixes are already in, except for some hard coded "post" type checks, looking into those now.

Issue #242, opened 2019-01-16
[bug] error calling Related: indexing currently not supported for index "keywords" and type []interface {}


pad92:
unable to build site, related page fail

but if I change into single template
```
        {{ if .Site.Params.showRelatedPosts }}
          {{ $related := .Site.RegularPages.Related . | first 3 }}
          {{ with $related }}
          <h4 class="see-also">{{ i18n "seeAlso" }}</h4>
          <ul>
          {{ range . }}
            <li><a href="{{ .RelPermalink }}">{{ .Title }}</a></li>
          {{ end }}
          </ul>
          {{ end }}
        {{ end }}
```
by

```
        {{ if .Site.Params.showRelatedPosts }}
          {{ range first 1 (where (where .Site.Pages ".Params.tags" "intersect" .Params.tags) "Permalink" "!=" .Permalink) }}
            {{ $.Scratch.Set "has_related" true }}
          {{ end }}

          {{ if $.Scratch.Get "has_related" }}
                  <h4 class="see-also">{{ i18n "seeAlso" }}</h4>
                  <ul>
                {{ $num_to_show := .Site.Params.related_content_limit | default 5 }}
                {{ range first $num_to_show (where (where .Site.Pages ".Params.tags" "intersect" .Params.tags) "Permalink" "!=" .Permalink) }}
                    <li><a href="{{ .RelPermalink }}">{{ .Title }}</a></li>
                {{ end }}
              </ul>

          {{ end }}
        {{ end }}
```
it's work, but i dont understand why it crash with the initial code version

NuLL3rr0r:
Thank you so much for your fix. I can confirm this bug on a multilingual site. If the website is not multilingual the related content works. But if it is, I get the same error as you. Applying your fix works. Thank you!

PR #243, opened 2019-01-16
[bugfix] related page 

pad92:
change for issue #242


halogenica:
Thank you for the fix!

PR #249, opened 2019-02-15
Update keybase icon

benz0li:
see https://fontawesome.com/icons/keybase?style=brands


halogenica:
Thank you for the PR!

PR #251, opened 2019-03-06
Added Kaggle to social icons

rishisubramanian:
Added Kaggle to the list of social media icons. Potentially useful for data scientists.


halogenica:
Thank you!

PR #255, opened 2019-03-20
Add KaTeX fonts for use with self-hosted option

soofaloofa:
The selfHosted option does not include the fonts required for KaTex. This PR includes those fonts so that KaTex works correctly when `selfHosted = true`.


halogenica:
Fantastic, thank you!

PR #264, opened 2019-04-20
Structured Data Fixes and Feature Image Support

DaveBben:
A couple of fixes giving better clarity for Structured data and adding support for featured image. Also adding rel="noopener" for links that have "target=_blank"


Ioan-Popovici:
How does the feature image support work?

henryiii:

Thanks @DaveBben, and sorry for the long wait. Most of this PR has landed over the years:

- `rel="noopener"` on the share links (now `noopener noreferrer`; the Google+ and StumbleUpon buttons were removed)
- `"@type": "Person"` on the article author, now built by `seo/structured/author-schema.html`
- `dateModified` from `.Lastmod` instead of `.Date`
- publisher name from the site title, now an `#organization` node

The one part still missing was the page-level image in the `Article` schema. That is #836, with you as co-author. I kept `share_img` and `image` as the parameter names rather than adding `featured_image`, to avoid a third name for the same value.

Closing this one, since the branch no longer applies to the current template layout.

PR #265, opened 2019-04-21
add missing .png pulled in by photoswipe.default-skin.min.css when self-hosting

NuLL3rr0r:
When Site.Params.selfHosted is on, if you click on an image photoswipe.default-skin.min.css pulls in static/css/default-skin.png and since it's not there you'll get a 404 error.

Steps to reproduce:

1. Open a beautifulhugo instance with Site.Params.selfHosted = true (local or remote).
2. Press F12 (I am using Firefox)
3. Click on Network tab
4. Refresh while Network tab is open
5. Click on an image so that PhotoSwipe loads the image
6. Look for the 404 error for default-skin.png in the Network tab


halogenica:
Thank you very much for the PR!

PR #270, opened 2019-04-26
Fixed error from HTML5 validator and used input type URL

VincentTam:
From the [HTML standard for `<textarea>`](https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element), the attribute `type` is *not* listed under "[**Content attributes**](https://html.spec.whatwg.org/multipage/dom.html#concept-element-attributes)".  As a consequence, passing _any_ site with [Staticman](https://staticman.net/) _enabled_ would reproduce this error in [W3C's HTML5 validator](https://validator.w3.org/nu/):

> **Error: Attribute `type` not allowed on element [`textarea`](https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element) at this point.**

### DIFF
--- a/layouts/partials/staticman-comments.html
+++ b/layouts/partials/staticman-comments.html
@@ -46,7 +46,7 @@
 
   <fieldset>
   <div class="textfield">
-    <textarea name="fields[comment]" type="text" placeholder="{{ i18n "useMarkdown" }}"></textarea>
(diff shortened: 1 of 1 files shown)


VincentTam:
See "Input type URL" in https://www.w3schools.com/html/html_form_input_types.asp.

halogenica:
Thank you so much!

PR #273, opened 2019-05-08
Adding last modified for pages if lastmod is set

henryiii:
This adds the "lastmod" information if it is set on a page. This is one of two changes I am always rebasing on top of the master branch (the other being the "figure -> beautifulfigure" rename so that I can use the built-in figure env, #169 - but that's a much larger (user-facing) change).


halogenica:
Thank you very much! I really appreciate your contribution :)

PR #276, opened 2019-05-20
Fixed staticman.css typo

VincentTam:
Found this while porting my PR for nested Staticman comments to [Huginn](https://framagit.org/staticman-gitlab-pages/huginn).  Btw, `:before` is CSS2 syntax, and its CSS3 counterpart is `::before`.

Reference: https://developer.mozilla.org/en-US/docs/Web/CSS/::before

### DIFF
--- a/static/css/staticman.css
+++ b/static/css/staticman.css
@@ -82,7 +82,7 @@
  	height: 10em;
 }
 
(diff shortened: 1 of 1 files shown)


halogenica:
Thank you very much for the fix!

PR #279, opened 2019-06-02
Remove duplicate meta tags

### LINKED ISSUE #277: Open Graph and Twitter meta tags are duplicated
The meta tags for Twitter and Open Graph are added by `main.html` from the `seo` folder as a partial in `head.html`, but they are also added directly in `head.html`, leading to many duplicated tags.

henribru:
Fixes #277


halogenica:
Thank you very much!

PR #287, opened 2019-06-16
Members section

cassiobatista:
## Purpose
This PR creates a section called "members" for the case of a page for a research or company group  instead of a personal one. It was mainly inspired by the Tags section, from which I borrowed the collapsible panels layout.

## Layout screen
out

## Details
- new section defined on `content/members/` with two Markdown files as examples representing two individual members
- new archetype `archetype/members.md` which contains in its own front matter every single social network defined in `data/*/social.toml`
- new social nets added to `data/beautifulhugo/social.toml`. They are mainly academic networks such as research gate, orcid, google scholar, etc. and I had to retrieve the icons from the Academicons (https://jpswalsh.github.io/academicons/) since they are not available on Font Awesome. The CSS is loaded at the `layouts/members/list.html` file. However it broke the footer of the member section (as you can see in the GIF above), and I haven't figured out a way to solve it :( Footers on the other pages are not affected, though.
- new section list layout defined in `layouts/members/` inspired on `layouts/_default/terms.html`
- new partial layout `layouts/partials/member-card.html` which is basically two columns to exhibit the member avatar and a short description side by side.
- `config.toml`file  altered to show the members section at the navbar
- `member.png` image added to `/static/img/` to serve as the default avatar for newly created members inside the content folder


Ioan-Popovici:
Excellent, thank you!
One little problem is that the main page social icons are a bit skewed.

cassiobatista:
Thanks :)

> One little problem is that the main page social icons are a bit skewed.

that's because of the academicons css file i've imported at the `layouts/member/list.html` file. I know almost nothing about css so I wasn't able to solve it :(

PR #292, opened 2019-07-14
Don't use google analytics when running on built-in server

nnja:
The google analytics code is inserted even when running the site on Hugo's [built-in development server](https://gohugo.io/commands/hugo_server/).

This wreaks havoc on analytics, especially with auto-reload turned on.

Instead, add a check to only add Google Analytics when the site **is not** running on Hugo built-in server.


halogenica:
I merged this in, just the minimal subset of code for the google analytics. Thanks!

PR #297, opened 2019-08-04
Correct links to hugo-easy-gallery.css.

hsn10:
Now this works with _hugo daemon_ using baseurl _/hugo_


VincentTam:
Hi.  Have you seen #169, which renames `layouts/shortcodes/figure.html` → `layouts/shortcodes/beautifulfigure.html`, so that this theme's `figure` won't override the behavior of Hugo's internal one?  Here's the feedback from halogenica, this theme's author:

> I definitely want to allow users to disambiguate between Hugo's figure shortcode and this theme's figure. Your change is probably the best way to do it, I'm just trying to think about breaking backwards comparability. I might just have to rip off the band aid :)

Excuse me if my question seems stupid: why have you put `absURL` instead of `relURL`?

hsn10:
I used absURL because loading other CSS files is done that way.

halogenica:
Thank you very much!

Issue #300, opened 2019-08-15
Hugo 0.57.0 breaks index/homepage (only Posts shown)


bobek:
Index page will only contain "Posts" link when built with hugo `v0.57.0`. It is caused by change of semantics of `.Pages` as noted on the release notes - https://github.com/gohugoio/hugo/releases/tag/v0.57.0

Fix is simple as bobek/beautifulhugo@79abb2e . But that is not backwards compatible.

bobek:
> Index page will only contain "Posts" link when built with hugo `v0.57.0`. It is caused by change of semantics of `.Pages` as noted on the release notes - https://github.com/gohugoio/hugo/releases/tag/v0.57.0
> 
> Fix is simple as [bobek/beautifulhugo@79abb2e](https://github.com/bobek/beautifulhugo/commit/79abb2e) . But that is not backwards compatible.

bobek:
I have just verified that `0.54.0` actually works with the patch as `RegularPages` collection was added long time ago (I have found them even in version `0.18.0` but version from last year don't work for other various reasons (e.g. partials)). So maybe this patch is actually acceptable as is.

henryiii:
I've found the same things and committed the exact same page independently. I highly recommend this gets merged!

mckmonster:
I did a pull request for that #302

halogenica:
This should now be fixed with https://github.com/halogenica/beautifulhugo/commit/1d75ccd689fd52f394077f5a5e4c88ca580a07bf - thank you for the report!

PR #301, opened 2019-08-16
Use Site.RegularPages for index

bobek:
Index page will only contain "Posts" link when built with hugo `v0.57.0`. It is caused by change of semantics of `.Pages` as noted on the release notes - https://github.com/gohugoio/hugo/releases/tag/v0.57.0

See issue #300

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -9,7 +9,7 @@
         {{ end }}
 
         <div class="posts-list">
-          {{ $pag := .Paginate (where .Data.Pages "Type" "post") }}
+          {{ $pag := .Paginate (where .Site.RegularPages "Type" "post") }}
           {{ range $pag.Pages }}
             <article class="post-preview">
               <a href="{{ .Permalink }}">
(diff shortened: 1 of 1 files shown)


layouts/index.html
```diff
@@ -9,7 +9,7 @@
         {{ end }}
 
         <div class="posts-list">
-          {{ $pag := .Paginate (where .Data.Pages "Type" "post") }}
+          {{ $pag := .Paginate (where .Site.RegularPages "Type" "post") }}
```
VincentTam: It's better to use [`mainSections`](https://gohugo.io/functions/where/#mainsections) to avoid hard-coding things.

bobek: Excellent suggestion @VincentTam . Can you squash commits when merging or do you want me create a new branch for you.

VincentTam: Oops.  I should have committed my changes on top of 79abb2e.  Thanks for changing your code.

henryiii:
This theme (along with many others) is broken in Hugo 0.57. This is because we are not using the "recommended" way to get pages. Highly recommend this patch! @halogenica

halogenica:
Thank you so much for the PR! I'm happy to accept it and I'm glad you @'d me to get my attention :)

PR #302, opened 2019-08-16
Fix paginator on index

mckmonster:
I had a bug on display of post list on the first page with the last Hugo version.

If you want I did this fix.

### DIFF
--- a/exampleSite/config.toml
+++ b/exampleSite/config.toml
@@ -28,6 +28,7 @@ pygmentsCodefencesGuessSyntax = true
   socialShare = true
   delayDisqus = true
   showRelatedPosts = true
+  mainSections = "post"
 #  gcse = "012345678901234567890:abcdefghijk" # Get your code from google.com/cse. Make sure to go to "Look and Feel" and change Layout to "Full Width" and Theme to "Classic"
 
 #[[Params.bigimg]]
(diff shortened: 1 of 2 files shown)


layouts/index.html
```diff
@@ -9,8 +9,7 @@
         {{ end }}
 
         <div class="posts-list">
-          {{ $pag := .Paginate (where .Data.Pages "Type" "post") }}
-          {{ range $pag.Pages }}
+          {{ range where .Paginator.Pages "Type" "post" }}
```
VincentTam: It's better to use `site.Params.mainSections` instead of hard-coded `"post"`.

mckmonster: I updated the code

exampleSite/config.toml
```diff
@@ -28,6 +28,7 @@ pygmentsCodefencesGuessSyntax = true
   socialShare = true
   delayDisqus = true
   showRelatedPosts = true
+  mainSections = "post"
```
onedrawingperday: It is not needed to enter mainSections in the config. Also the way you have entered this parameter is wrong since it is an array not a string.

Please remove this.

mckmonster:
@onedrawingperday it's for an exemple.
It's better if this line is in exemple.
If it's not the case, the exemple doesn't display anything.

bobek:
Thanks @mckmonster , I would consider it duplicate of #301 .

mckmonster:
Yes, you're right @bobek.
We did it at the same times.
But I made more change on buttons display for "prev" and "next" page

halogenica:
Thank you for the change! I've merged #301 so I'm going to close this since I believe it's mostly a duplicate.

PR #303, opened 2019-08-31
Added video shortcode

rinajhinjhina:
Added a video shortcode based on the figure shortcode. It is something I use quite regularly and wanted to provide it here.


RoneoOrg:
Hi @rohanfaiyazkhan,

adding this feature would be more relevant with a reference in the documentation, would you like to provide an update to the Readme?

PR #305, opened 2019-09-08
To have the same display on each pages

mckmonster:
I made a change to display post-preview same on all pages (index and list)


halogenica:
Thank you!

PR #225, opened 2018-11-20
[Feature] Archives

pad92:
add archive page for #220

PR #307, opened 2019-09-21
Revert "[Feature] Archives"

halogenica:
Reverts halogenica/beautifulhugo#225


halogenica:
I like this feature and want the functionality, but I don't like that a section defaults to the archive view (and that the page is titled "archive" instead of the section). I need to think through this one a little more before merging back.

milosgajdos83:
Ah, this is a shame. I kind of liked this and now miss it :-)

milosgajdos:
Just wondering, are you still thinking of bringing this back, or have you now completely discarded the idea?

PR #275, opened 2019-05-14
Add Utteranc

yongju-jin:
Add Utteranc for comments.
https://utteranc.es/

PR #308, opened 2019-10-06
Utterances

jklaise:
Cleaned up #275 . Just integrated this on my own blog, works like a charm!

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ Add *Utterances* configuration section in `config.toml` or `config.yaml`
 Sample `config.toml` configuration
 
 ```
-[Params.Utteranc]
+[Params.Utterances]
   repo = "user/repo"
   issue_term = "pathname"
   label = "comment"
--- a/README.md
+++ b/README.md
@@ -106,6 +106,20 @@ comments:
 
 If you *don't* have the section `[Params.staticman]` in `config.toml`, you *won't* need the section `reCaptcha`  in `staticman.yml`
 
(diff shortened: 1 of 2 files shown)


jklaise:
On second thought the script should probably go in `header.html` with the rest of the links/stylesheets.

VincentTam:
> On second thought the script should probably go in `header.html` with the rest of the links/stylesheets.

Why don't load the script for Utterances just like Disqus?

jklaise:
@VincentTam I'm not that familiar with this, the [Utterances website](https://utteranc.es/) shows the script embedded as is (complete with the `async` parameter). What would need to change to make it load similar to Disqus loading? I'm looking at `header.html` for Disqus but can't work out what's happening there.

VincentTam:
> @VincentTam I'm not that familiar with this, the [Utterances website](https://utteranc.es/) shows the script embedded as is (complete with the `async` parameter). What would need to change to make it load similar to Disqus loading? I'm looking at `single.html` for Disqus but can't work out what's happening there.
> 
> Edit: it looks like Disqus is loaded (delayed) in `footer.html`?

I'm familiar with neither one.  I mean to load either one in the same Go-HTML template, like what you've already done.

https://github.com/halogenica/beautifulhugo/blob/01b72f4fe351704cee10cf36a8c31d2eafa53a38/layouts/_default/single.html#L78-L93

jlozadad:
is there a similar tool for gitlab?

PR #311, opened 2019-11-02
Fully remove the avatar image when scrolling down

fri-sch:
Currently the opacity of the avatar/logo image is faded to zero on
scrolling down, but the element is still clickable. To not make it
only invisible, but also completely remove it, we also animate the
visibility property.


halogenica:
Awesome, thank you very much!

PR #312, opened 2019-11-06
Add korean i18n module

Pickhealer:
There is no i18n module for korean. So I add ko.yaml in i18n.


halogenica:
Wonderful, thank you so much!

Issue #194, opened 2018-09-24
Support for "dark mode" in macOS Mojave


igorkulman:
Maybe just switching the text and background color would look fine.

Should be detectable with `prefers-dark-interface` media query according to [How to detect if OS X is in dark mode in browsers?](https://stackoverflow.com/a/51799496/581164).

halogenica:
Thanks for the link on how to detect this! It's unlikely that I'll make a dark option for this theme any time soon, but I'd be happy to accept a PR if someone wants to give it a shot.

henryiii:
I've have a dark mode patch running here: https://iscinumpy.gitlab.io. It's a single commit, I can make it into a PR if you want it.

henryiii:
(Note: iOS will be released on Friday with dark-mode support, followed by iPadOS a week or two later, so this will start affecting more people soon)

henryiii:
Commit is here: https://github.com/henryiii/beautifulhugo/commit/f64b4735bc6811890a877f388ba3d6c7a00a216e

ibnunowshad:
Was this commit PR has been completed? I am willing to test this.

henryiii:
I need to sync it with my copy, I think I’ve polished it a bit over time.

ibnunowshad:
> I need to sync it with my copy, I think I’ve polished it a bit over time.
I wait for your updates. I am eagerly waiting for this update. I am just curious that this feature support for macOS Catalina `auto` mode. So `light` and `dark` themes based on macOS settings.

henryiii:
Made the PR, @halogenica will need to approve. I've been using it on iscinumpy.gitlab.io for a while now; have had to adjust a few colors at first, but pretty happy with it.

ottok:
The PR#317 was merged. If the commit https://github.com/halogenica/beautifulhugo/commit/8043892164058b207f38fd5ec5c899b01722ffb8 would have had in the body the text `Closes. #194` this issue would have been closed automatically on merge..

henryiii:
Full dark mode with colorScheme param and toggle button implemented

PR #317, opened 2019-11-29
Dark mode support

henryiii:
Dark mode support (see #194).

You can see an example on the right here (slightly older version):

https://twitter.com/HenrySchreiner3/status/1174691026650521602/video/1


static/css/main.css
```diff
@@ -95,30 +117,27 @@ div.panel-body a.list-group-item.view-all {
...
-  color: white;
-  background: transparent;
-}
-img::-moz-selection {
+img::selection, img::-moz-selection {
   color: white;
   background: transparent;
 }
 
+
```
julian-alarcon, line 130: You can remove this @henryiii

fubarhouse:
I've added a `custom.css` to implement this, thanks for your efforts @henryiii.
FWIW it works well, so if anybody wants testers you have my opinion.

rterakedis:
I've also implemented these changes on my site with success:   https://blog.euc-rt.me

Thanks @henryiii

halogenica:
Thank you very much @henryiii !

PR #168, opened 2018-05-07
Adding some useful shortcodes

henryiii:
This adds two useful shortcodes:
* details: an expandable list supported on most (all modern) browsers
* columns: A side by side comparison environment, supporting narrow windows (like iPhones) gracefully.

Notes added to readme about the new shortcodes.

<details><summary>If you'd like to see more about details, click here</summary>
This is what it looks like.

😄 
</details></br>

Note: nested shortcodes are not used for columns due to a bug in Hugo: https://github.com/gohugoio/hugo/issues/1642

PR #323, opened 2019-12-11
Fix details for Hugo 0.60+

henryiii:
The markdown parser now strips raw html (since Hugo 0.60), so `{{<>}}` is required for all shortcodes containing HTML. This will cause the details shortcode to start failing to render markdown if the change is made. This fixes it to continue to render markdown inside.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,9 +158,9 @@ There are two extra shortcodes provided (along with the customized figure shortc
 This simply adds the html5 detail attribute, supported on all *modern* browsers. Use it like this:
 
 ```
-{{% details "This is the details title (click to expand)" %}}
+{{< details "This is the details title (click to expand)" >}}
 This is the content (hidden until clicked).
-{{% /details %}}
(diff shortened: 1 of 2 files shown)


henryiii:
@halogenica, any chance this fix can go in? I was the original author of the details shortcode in #168.

akiradeveloper:
I am sorry. I see some inconsistency between what you said and what you did. What you did is just piping the value into markdownify function not surrounding something with the brackets.

henryiii:
The fix is to manually ensure that markdownify is called in the function (this PR), and then for users to use `{{< details "..." >}}` instead of `{{% details "..." %}}`. It really doesn't change much if `{{%` is used; 0.60+ will not render it properly. However, `{{<` will work, and is the correct path going forward.

(If there's an example somewhere, that should be fixed here too, didn't look for it)

henryiii:
Thanks! I fixed the example.

PR #325, opened 2019-12-11
Display summary in post preview

### LINKED ISSUE #324: Summary not being displayed in post preview
#171 is not quite fixed. If a summary exits in the front matter the full content is shown and not the summary. 

Following up on #171 , if the author has specified a summary, this should be shown instead of the truncated (mostly malformed) text.

YaguraStation:
fixes #324

### DIFF
--- a/layouts/partials/post_preview.html
+++ b/layouts/partials/post_preview.html
@@ -21,10 +21,13 @@
     </p>
     <div class="post-entry">
         {{ if .Summary }}
-        {{ .Summary }}
-        <a href="{{ .Permalink }}" class="post-read-more">[{{ i18n "readMore" }}]</a>
+            {{ .Summary }}
         {{ else }}
-        {{ .Content }}
+            {{ if .Truncated }}
+            <a href="{{ .Permalink }}" class="post-read-more">[{{ i18n "readMore" }}]</a>
+            {{ else }}
+            {{ .Content }}
+            {{ end }}
         {{ end }}
     </div>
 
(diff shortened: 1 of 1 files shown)


reesehopkins:
@YaguraStation thanks! This worked for me.

akiradeveloper:
@YaguraStation this works for me too.

To push the merge of this PR I will share my experience:
I have a blog in Japanese and since hugo doesn't count the Japanese words correctly (it tends to be small number) the automatic summary split always go full content. Then two options remain: more splitter and front matter. The more splitter does work without this PR but the summary in front matter doesn't work as mentioned. Since my blog is not a research paper that always have summary at beginning, I prefer front matter way. That is why I need this change.

layouts/partials/post_preview.html
```diff
@@ -20,11 +20,14 @@ <h3 class="post-subtitle">
         {{ partial "post_meta.html" . }}
     </p>
     <div class="post-entry">
-        {{ if .Truncated }}
-        {{ .Summary }}
-        <a href="{{ .Permalink }}" class="post-read-more">[{{ i18n "readMore" }}]</a>
+        {{ if .Summary }}
+            {{ .Summary }}
```
grisu48, line 24: Hey there! Thanks for the PR, this indeed fixed the issue also for me. But when defining a summary, now the [Read More] link is missing, so I suggest to add it here as well:

```suggestion
            {{ .Summary }}
            <a href="{{ .Permalink }}" class="post-read-more">[{{ i18n "readMore" }}]</a>
```

PR #326, opened 2019-12-25
Fix dateFormat translation for RU

armatusmiles:
Hello. I use two languages for my blog. But in Russian translate date shows incorrectly.
This change is helping me to fix bug

The incorrect "translated" date. ( Current master ) 
As you can see here the all months is "Январь" but it is not a real date.
date

The correct date ( After this changes ) 
date2


halogenica:
Thank you so much for your fix!

PR #332, opened 2020-01-14
Change Twitter Card from summary to summary_large_image

sethiele:
With this change you get the bigger [Twitter Card Preview](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/summary-card-with-large-image)


halogenica:
Thank you!

PR #333, opened 2020-01-14
Let Hugo resize the Post Preview Image in Overviews

sethiele:
For a smaller Post Preview Image I changed the way of adding it to the post_overwiew.
It use the Hugo [Resize](https://gohugo.io/content-management/image-processing/#resize) function to fit it to the max size of image withe. 
Therefore the Image have to be stored in the post folder with `*_header.{png,jpg}`as name.

The old way with the page variable will also work, but without resizing.

### DIFF
--- a/layouts/partials/post_preview.html
+++ b/layouts/partials/post_preview.html
@@ -6,8 +6,15 @@
             {{ .Params.subtitle }}
         </h3>
         {{ end }}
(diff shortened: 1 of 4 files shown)


henryiii:
This is mostly possible now in master, minus the one auto-detect feature, which I think is better handled other ways today. Thanks though!

PR #336, opened 2020-02-21
WIP: use site.Params.mainSections instead of "page"

vdot0x23:


vdot0x23:
A similiar pattern with "post"/"page" is used in single, header, and footer.

henryiii:
I think this was fixed in #692.

PR #337, opened 2020-02-26
use page leaf bundle image resources with figure shortcode

wtip:
This change allows you to use [page leaf bundle](https://gohugo.io/content-management/page-bundles/#leaf-bundles) image resources with the `< figure >` shortcode. This means you can use images stored with the post in a directory instead of having to store all of your images in the root static directory.

Additional changes would be needed to support directory galleries. They might be here https://github.com/liwenyip/hugo-easy-gallery/issues/54 but I have no need for this so I didn't spend any time reviewing it.


henryiii:
Thanks, but I think this is now handed via page-bundle.

Superseded by the image-resource work (#822, #833 and the beautifulfigure shortcode), which supports page-bundle and assets/ images while keeping the static-path fallback.

PR #341, opened 2020-04-26
Gallery: Fix lazy loading if thumbs are present

flo333:
Specify image sizes at build time, so load-photoswipe.js doesn't have to get all full-size images on page load.


henryiii:

Thanks! The page-bundle gallery path already emits sizes, and the `static/` path now does too as of 1613236, using `imageConfig` as proposed here. Closing since this branch no longer applies to master.

PR #343, opened 2020-05-31
Translation in Lombard

bsciretti:


halogenica:
I love seeing a new translation, thank you so much!

PR #346, opened 2020-06-25
Enable link verification per https://docs.joinmastodon.org/user/profile/

utahcon:
This change adds `rel="me"` to the mastodon icon link in the social bar. This enables Mastodon to verify your blog/account as owned by the same person.


halogenica:
Thank you!

PR #348, opened 2020-07-22
Raw html in the content files

astonbitecode:
Hi. 

As far as I understand, after Hugo version 0.62 we cannot add raw html in the markdown files. 

Would you like to add a shortcode that allows this? 

Something like:

```
{{< rawhtml >}}
  <p class="speshal-fancy-custom">
    This is <strong>raw HTML</strong>, inside Markdown.
  </p>
{{< /rawhtml >}}
```

The idea is based on this [blog](https://anaulin.org/blog/hugo-raw-html-shortcode/).


RoneoOrg:
Hi @astonbitecode,

Hugo allows to use HTML in content files, even with a version superior to 0.60, with the following settings in `config.toml`:

```yaml
    [markup.goldmark.renderer]
      unsafe = true
```
Ref:
- https://gohugo.io/getting-started/configuration-markup#goldmark
- https://discourse.gohugo.io/t/hugo-cms-and-unsafe-true/35069

Should we still consider this PR or does this setting answer your needs?

astonbitecode:
This seems enough.

Thanks!

Issue #349, opened 2020-07-25
Post preview tag link has double forward slash on multilanguage website


dovidio:
Hi, first of all thanks for this wonderful theme!

I have a multilanguage website, with some default language, something like this
```
baseURL = "/"
defaultContentLanguage = "en"
```
When I'm on the page listing post previews, all the posts with tags have a wrong link with double slash `//tags/:tagname`  which results in a 404 when clicked.
I fixed it on my website in this [commit](https://github.com/dovidio/personalwebsite/commit/34762e94c29fd2c26c16c45f8ae2de21bdf9b46d)
I'm not sure if this would break other configurations. I can create a PR if needed

jvincentnz:
@dovidio, encountered this issue and used your [fix](https://github.com/dovidio/personalwebsite/commit/34762e94c29fd2c26c16c45f8ae2de21bdf9b46d) for `layouts/partials/post_preview.html`. I also needed to apply your fix to `layouts/_default/single.htm` to resolve all occurrences of the tag link having double forward slashes on my website.

See my PR #354 for the required changes.

dovidio:
@jvincentnz You're right, I have overlooked that. Hopefully you're changed will get merged soon 🤞

marcosimbuerger:
@halogenica can we merge PR https://github.com/halogenica/beautifulhugo/pull/354? Thank you!

PR #354, opened 2020-09-12
Fixes Upstream Issue 349 - tag link has double forward slash

jvincentnz:
Fix for "https://github.com/halogenica/beautifulhugo/issues/349" where "tags" have double forward slashes causing broken links.
Takes the fix from "https://github.com/dovidio/personalwebsite/commit/34762e94c29fd2c26c16c45f8ae2de21bdf9b46d" for "layouts/partials/post_preview.html" and extends it to also fix "layouts/_default/single.html".

### DIFF
--- a/layouts/_default/single.html
+++ b/layouts/_default/single.html
@@ -8,7 +8,11 @@
         {{ if .Params.tags }}
           <div class="blog-tags">
             {{ range .Params.tags }}
+              <!-- Fix for "https://github.com/halogenica/beautifulhugo/issues/349".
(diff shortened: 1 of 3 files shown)


jvincentnz:
Updated `terms.html` layout file to fix the extra forward slash as per @nelsonwellswku.

henryiii:
Thanks!

henryiii:
I’ll probably delete the commented version at some point if I get around to it, I think this is the correct way to do it so I don’t need a commented bad way to do it. :) Having a note is fine, just not the commented out code.

ibnunowshad:
Will this fix multilanguage such as en, fr for slashes?

PR #363, opened 2020-11-08
Correct Italian translation

pfmaggi:
As a native language speaker I think that this improves the current translation in Italian.


halogenica:
Thanks!

PR #367, opened 2021-01-21
Implement CSS minification

RoneoOrg:
Hugo Pipes allow [the minification of CSS assets](https://gohugo.io/hugo-pipes/minification/)

This branch is an implementation of this feature.

**Key advantages:**

- every single CSS file is now minified
- half of them are also concatenated
- as the minification is done on the fly, previously minified files are now shipped [unminified](https://www.unminify2.com/). The CSS files are easier to analyze and customize

The default behavior is set to use selfHosted files instead of CDNs, as it's required to take advantage of the minification.

**Carefully tested with:** Hugo 0.56 _basic_ ,  Hugo 0.80.0 _basic_ and  Hugo 0.80.0 _extended_


RoneoOrg:
The CSS used in shortcodes in now minified too

RoneoOrg:
TODO: Potential improvement: make the minification optional via a Param in `config.toml`

henryiii:
You can use `--minify` now (and css has been redesigned).

PR #371, opened 2021-01-21
Responsive postmeta

RoneoOrg:
The post meta could break better on small screens with a bit of Flexbox


rouke-broersma:
This messes up my page on a 1080p desktop monitor.

RoneoOrg:
Hi @rouke-broersma and thanks for the report,
could you provide a link to check the issue you're facing?

rouke-broersma:
I've reverted the change on my fork so no, but the issue is that the summary is pushed to the side of the post block when the post has a cover image. It's squashed between the normal post width and the right border of the screen.

Bas-Man:
This might be able to be closed as the theme has changed significantly.

PR #372, opened 2021-01-22
Don't render header if title is empty

briantoth:
image

This is the current behavior on a small screen if the title is empty.

image

After the change


halogenica:
Thanks!

PR #373, opened 2021-01-22
fix typos in translation files

RoneoOrg:
Fix two typos in translation files (yes, "address" is written "adresse" in french, with only one `d`)


halogenica:
Thank you!

PR #380, opened 2021-02-16
Use jQuery slim

### LINKED ISSUE #342: bump jquery version in partials/footer file
I saw it is loading 1.12.4 which is from 2016 and has Vulnerabilities. 

Is the JQuery library anywhere used? is it used in staticman? if not remove it or load only when staticman enabled?

VincentTam:
Resolves #342.

Beautiful Jekyll from which this theme was ported switched to the slim version, which doesn't include the `$.ajax()` function.  I changed the Staticman JS script in the same way as my PR daattali/beautiful-jekyll#782 to make it compatible with the slim version.


halogenica:
Thank you!

PR #382, opened 2021-02-17
Display inline math using the '$' delimiter

VincentTam:
That's the default LaTeX syntax.  Using `\\(` and `\\)` worsens the editing experience.

Code extracted from https://github.com/KaTeX/KaTeX/blob/5c44c47bd2ca70c2a8585c19d9fae598b1227a57/contrib/auto-render/index.html#L44-L48

~I'm not going to repeat the `delimiters` found in `static/js/auto-render.min.js`.~


henryiii:
You can use `$$` (now?). A single $ would be to easy to trigger by mistake, I think.

PR #387, opened 2021-03-03
Corrected TR translation

tunix:
Inside the EN translation, these variables are defined as `{{ . }}` -- but in Turkish translation, it was `{{ .Count }}` which seems to be incorrect and causes empty parts inside the rendered HTML.


halogenica:
Thank you for the fix!

PR #388, opened 2021-03-10
Adding Utterances Comments Support

rterakedis:
Adding all the pieces to use Utterances for commenting system.


rterakedis:
Accidentally added a merge for disclaimer text in the footer in this Pull Request also.

PR #395, opened 2021-06-26
Added glyphicons files to fonts folder.

sunpech:
Fixes glyphicons not showing in selfHosted set to true.


sunpech:
These are the glyphicon files from bootstrap-3.3.7.

henryiii:
Bootstrap 5 is used now.

Issue #397, opened 2021-07-01
Duplicated title


alienbogart:
Right now, on the main page, beautifulhugo sets the title as "Page Title - Page Title". I wish to avoid the repetition. How can I do that?

whatacold:
I can confirm this, I find the `<title>` code at `head.html` but don't know how to fix it yet.

RoneoOrg:
In `layout/partials/head.html`, replace

```
<!-- Site Title, Description, Author, and Favicon -->
{{- with ($.Scratch.Get "Title") }}
  <title>{{ . }} - {{ $.Site.Title }}</title>
{{- end }}
```

with

```
<!-- Site Title, Description, Author, and Favicon -->
  {{ if .IsHome }}
    {{- with ($.Scratch.Get "Title") }}
      <title>{{ . }}</title>
    {{- end }}
  {{ else }}
    {{- with ($.Scratch.Get "Title") }}
      <title>{{ . }} - {{ $.Site.Title }}</title>
    {{- end }}
  {{ end }}
```

This should normally do the trick

RoneoOrg:
A pull request is now pending for review: https://github.com/halogenica/beautifulhugo/pull/398

PR #398, opened 2021-07-11
Remove duplicate Title on Homepage

RoneoOrg:
Fixes https://github.com/halogenica/beautifulhugo/issues/397

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -27,9 +27,16 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
 <!-- Site Title, Description, Author, and Favicon -->
-{{- with ($.Scratch.Get "Title") }}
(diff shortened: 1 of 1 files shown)


whatacold:
Thanks for this quick fix, would you like to also respect `.Site.Params.homeTitle`, as in `header.html`?

BTW, what's the difference between `$.Scratch.Set "Title"` and `$.Scratch.Set "title"`? I'm confused.

RoneoOrg:
Hi @whatacold ,

> would you like to also respect .Site.Params.homeTitle, as in header.html?

Of course that's possible. It seems like this solution works fine though, could you please elaborate?

> BTW, what's the difference between $.Scratch.Set "Title" and $.Scratch.Set "title"?

I'm not perfectly sure of the right answer here, what about raising this question on discourse.gohugo.io?

RoneoOrg:
Hi @whatacold,

I just submitted a PR fixing this issue, and using `.Site.Params.homeTitle`
Please review and comment: https://github.com/halogenica/beautifulhugo/pull/444
Thanks!

/cc @shotakaha @fcapano

PR #402, opened 2021-07-14
fix: photoswipe doesn't work for <figure> without <a> inside.

whatacold:
Hi,

I use hugo with Emacs orgmode, the html doesn't have a `<a>` inside a `<figure>`, it looks like (or you can take a look at my this post: https://whatacold.io/blog/2021-07-11-babashka-random-password-generator/):
```html
<figure>
<img src="/img/2021-07-11-babashka-random-password-generator.png" alt="/img/2021-07-11-babashka-random-password-generator.png" title="/img/2021-07-11-babashka-random-password-generator.png" /><figcaption>
Generate Random Passwords in No Time
</figcaption>
</figure>
```

PhotoSwipe doesn't work with that snippet.

### DIFF
--- a/static/js/load-photoswipe.js
+++ b/static/js/load-photoswipe.js
@@ -16,11 +16,11 @@ $( document ).ready(function() {
 		var $figure = $(this),
 			$a 		= $figure.find('a'),
 			$img 	= $figure.find('img'),
-			$src	= $a.attr('href'),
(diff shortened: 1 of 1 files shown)


henryiii (approved):

whatacold:
thanks!

PR #412, opened 2021-09-08
Add Social Icons.

guilt:
Hi,

I was missing some social icons when I was trying this theme out, and decided to add some. Please
incorporate them if you like them.


ibnunowshad:
What does absolute URL thing here? Pardon my ignorance. Was that relate to removing language path?

guilt:
Yes, I tried doing the relURL but the tags didn't work for me either when I tried hosting it. This fix is what worked for me.

ibnunowshad:
@guilt May you please post some screenshot of before and after relate to _relURL_ and _absolute URL_?

guilt:
@ibnunowshad that fix is no more necessary (absURL). This only has the added social icons now.

henryiii (approved):

Issue #426, opened 2022-05-23
Hand written summary= is not used


hsn10:
If you use hand written summary= its not used, instead content is rendered

PR #427, opened 2022-05-23
Use hand written summary and make link to read rest

hsn10:
Fix for #426

### DIFF
--- a/layouts/partials/post_preview.html
+++ b/layouts/partials/post_preview.html
@@ -20,7 +20,7 @@
         {{ partial "post_meta.html" . }}
     </p>
     <div class="post-entry">
-        {{ if .Truncated }}
+        {{ if or (.Truncated) (.Params.summary) }}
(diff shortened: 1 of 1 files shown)


JonasVautherin (approved):
Works for me :+1:

JonasVautherin:
Since @henryiii is active now, I'd like to ping on this one.

To me it really fixes the bug, otherwise the home page does not render correctly (both the summary and the page are rendered together, instead of only the summary)

danielwagn3r:
Is there any plan to merge this? Summaries with Hugo have some problems in general, i.e. when it comes to images, code-snippets, ..., and [Front Matter Summary](https://gohugo.io/content-management/summaries/#front-matter-summary) is for sure a way to help out.

henryiii:
This looks reasonable. Feel free to ping, I'll try to spend some time on this in the future, but have a long list of things to finish first. :)

PR #430, opened 2022-05-30
Solves compatibility issues with Google Analytics v4

dgasparri:
See Hugo's internal template page https://gohugo.io/templates/internal/:

"When using Google Analytics v4 use _internal/google_analytics.html."


halogenica:
Thank you!

Issue #322, opened 2019-12-04
Markdown does not render in figure shortcode


ahelwer:
As discussed in gohugoio/hugo#6580, the custom figure shortcode does not render markdown in its caption, as it should per [official documentation](https://gohugo.io/content-management/shortcodes/#figure). Issue is also opened against GitLab Pages Hugo example repo, which will have to update its theme once this issue is resolved: https://gitlab.com/pages/hugo/issues/43

ahelwer:
Closing since there doesn't seem to be interest in fixing this and it's pretty easy to work around.

PR #435, opened 2022-07-15
Fix figure linkattr parameter

### LINKED ISSUE #258: figure attr and attrlink broken
The {{< figure ... >}} supports attr and attrlink parameters to link back to the original image. The beautifulhugo implementation of this is currently broken. Attempting to set `attrlink` results in:

```
themes/beautifulhugo/layouts/shortcodes/figure.html:23:58": execute of template failed: template: shortcodes/figure.html:23:58: executing "shortcodes/figure.html" at <.Get>: can't evaluate field Get in type string
```

ahelwer:
Have to use `$` to operate from the root scope inside a `with`
Closes #258

### DIFF
--- a/layouts/shortcodes/figure.html
+++ b/layouts/shortcodes/figure.html
@@ -20,7 +20,7 @@ Documentation and licence at https://github.com/liwenyip/hugo-easy-gallery/
         {{- if or (.Get "caption") (.Get "attr")}}
           <p>
             {{- .Get "caption" -}}
-            {{- with .Get "attrlink"}}<a href="{{.}}">{{ .Get "attr" }}</a>{{ else }}{{ .Get "attr"}}{{ end -}}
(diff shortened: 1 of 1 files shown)


henryiii (approved):
This looks correct to me.

(Not related) Personally, I'd like to make the an optional extension and not replace the default figure. (I've got an issue on that somewhere)

ahelwer:
Yeah I actually use a fork of this theme where the only change is that this figure shortcode is deleted. Some problems I've found with it:

- Markdown in caption isn't rendered #322 
- Very weird method of rendering figures that contain both `caption` and `attr`, especially if `attrlink` is present
- Doesn't respect figure sizing parameters like `width`
- If you click the figure it pops up in this weird way where it isn't even really zoomed in, which also breaks the `link` attribute; also works terribly if the figure is a vector diagram with a transparent background

So you've got my vote to get rid of it or make it optional.

ahelwer:
Closing since there doesn't seem to be interest in merging this and it has been superseded

PR #438, opened 2022-08-25
Fix figure shortcode with attrlink

### LINKED ISSUE #258: figure attr and attrlink broken

kmorel:
There were a couple of problems when using the custom figure
shortcode with the attrlink parameter. First, there was a
scoping issue that caused the hugo templates to fail when
attempting to get the attr parameter. Second, the css for
the anchor tag placed the link in the wrong place.

This is similar to PR #435 except that it also fixes a secondary problem with the css for the generated anchor tag.

Closes #258

### DIFF
--- a/layouts/shortcodes/figure.html
+++ b/layouts/shortcodes/figure.html
@@ -20,7 +20,7 @@ Documentation and licence at https://github.com/liwenyip/hugo-easy-gallery/
         {{- if or (.Get "caption") (.Get "attr")}}
           <p>
             {{- .Get "caption" -}}
-            {{- with .Get "attrlink"}}<a href="{{.}}">{{ .Get "attr" }}</a>{{ else }}{{ .Get "attr"}}{{ end -}}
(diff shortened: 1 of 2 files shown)


kmorel:
Well, I thought I had the attrlinks working, but there is still an issue. This change allows the link to be applied and places the text in the right place of the caption, but I just realized that the link does not work. For reasons I don't understand, the link for the image seems to extend over the figcaption tag, so when you click the attribute link, you actually get the image link overlayed on top of it.

I don't really understand this behavior, and I don't know how to fix it.

henryiii:
I think this was fixed in #655! Thanks!

PR #439, opened 2022-09-02
Add demo to the Readme

RoneoOrg:


RoneoOrg:
Merging this minor update pending for a week

henryiii:
Why not put this in GHA + GH Pages? Then the building code is public. I'd be happy to [set that up](https://scikit-hep.org/developer/gha_basic#github-pages), but it does need @halogenica to turn on the new publish to GH Pages setting in settings -> pages.

(I"m not very available till after next week though; teaching a class and speaking at two overlapping workshops).

RoneoOrg:
> Why not put this in GHA + GH Pages? Then the building code is public. I'd be happy to [set that up](https://scikit-hep.org/developer/gha_basic#github-pages), but it does need @halogenica to turn on the new publish to GH Pages setting in settings -> pages.

Indeed I do not have access to the Pages settings, let's wait for halogenica

RoneoOrg:
> Why not put this in GHA + GH Pages? Then the building code is public.

Another advantage: now the build is done from a separate repo, which I have to sync manually. Using a build right from this repo would ease the process

halogenica:
Hey @henryiii @RoneoOrg this GHA is new to me. I just enabled it and the demo site is being served here: https://halogenica.github.io/beautifulhugo/ (which redirects to halogenica.net/beautifulhugo, but it's probably best to use https://halogenica.github.io/beautifulhugo/ when linking elsewhere in case I change my custom domain name at some point). I don't know how to set up the GH Action to actually run Hugo, is that something you can configure now that I've enabled the Pages functionality? Thank you so much for working on this and let me know if you need me to do anything else!

henryiii:
Sure, I can set it up. In fact, Hugo is one of the [starter workflows](https://github.com/actions/starter-workflows/blob/main/pages/hugo.yml). :) Did you turn on the Actions source for Pages?

[Pages selector](https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta/)

PR #447, opened 2022-09-24
Implement more general Gitlab link.

### LINKED ISSUE #436: Make the Gitlab icon link more general
One nice advantage of gitlab, compared to github, is the possibility of hosting it somewhere else than gitlab.com, but right now in beautifulhugo you can't link the social icon to any other instance of gitlab.

**Suggestion**  
To make it possible, in the theme file `themes/beautifulhugo/data/beautifulhugo/social.toml`, for the base gitlab URL to be different from the default `https://gitlab.com/`, in order to permit linking to instances hosted elsewhere. The simplest way to achieve this is probably by changing `url = "https://gitlab.com/%s"` to `url = "%s"`. Leaving it to the user to provide the full URL.

It's rather easy to implement and would be a welcome improvement.

stafusa:
Closes #436


henryiii:
Would it be possible to detect if the host was missing and include a default? This will break all existing websites linking to gitlab.

stafusa:
> Would it be possible to detect if the host was missing and include a default? This will break all existing websites linking to gitlab.

True, allowing different hosts is probably not important enough to break current configurations. At the same time I wouldn't want to make Gitlab a special case with dedicated HTML code and new variables.

What do you think of introducing "gitlabalt", that takes "server/username" (like my current suggestion), and bringing back "gitlab" the way as it is in main, with gitlab.com implicit?

henryiii:
I don't think that scales to other uses very well. GitHub also allows custom host URLs, so it would also be nice there. Where is this being consumed? I think we should first verify it's not easy just to detect a slash and add a default if it's not present.

stafusa:
> Edit: as a more general solution, I'd say a full URL here ideally would render as a full URL. The final consumer is
> 
> https://github.com/halogenica/beautifulhugo/blob/4b5a7c7534c9acdb3e185adc6b23318c91c535dd/layouts/partials/footer.html#L12
> 
> So if that just checked to see if it started with `https://`, then used it directly, wouldn't that be a more general solution that would also work for GitHub and others?

That's definitely a proper way of doing it @henryiii . Thanks for the pointer. Please check whether I got something wrong in my latest implementation. It tests for "http[s]://" as you suggested.

henryiii (approved):
LGTM! Will wait a day or so to see if any other maintainers have input, then will merge. (remind me if I forget!)

stafusa:
> LGTM! Will wait a day or so to see if any other maintainers have input, then will merge. (remind me if I forget!)

@henryiii Your friendly reminder. :)

PR #448, opened 2022-09-29
ci: add GHA demo build

henryiii:
See #439. Pretty sure Pages has not be set to use custom Actions as a source (since it seems to being trying to build without intervention), that will need to be set in settings. settings -> pages -> source. CC @halogenica.

### DIFF
--- a/.github/workflows/hugo.yml
+++ b/.github/workflows/hugo.yml
@@ -30,7 +30,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      HUGO_VERSION: 0.102.3
+      HUGO_VERSION: 0.104.2
     steps:
       - name: Install Hugo CLI
         run: |
--- a/.github/workflows/hugo.yml
+++ b/.github/workflows/hugo.yml
@@ -1,0 +1,70 @@
+# Based on the sample workflow for building and deploying a Hugo site to GitHub Pages
+name: Deploy Hugo site to Pages
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["master"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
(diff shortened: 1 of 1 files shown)


.github/workflows/hugo.yml
```diff
@@ -0,0 +1,70 @@
...
+
+# Allow one concurrent deployment
+concurrency:
+  group: "pages"
+  cancel-in-progress: true
+
+# Default to bash
+defaults:
+  run:
+    shell: bash
```
henryiii, line 26: Not sure why this is here in the default workflow. bash is the default on linux. Otherwise the starter workflow looks pretty good.

.github/workflows/hugo.yml
```diff
@@ -0,0 +1,70 @@
...
+
+jobs:
+  # Build job
+  build:
+    runs-on: ubuntu-latest
+    env:
+      HUGO_VERSION: 0.102.3
```
henryiii, line 33:
```suggestion
      HUGO_VERSION: 0.104.2
```

Template's a bit out of date.

.github/workflows/hugo.yml
henryiii:
```suggestion
```

.github/workflows/hugo.yml
```diff
@@ -0,0 +1,66 @@
...
+  # Deployment job
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Deploy to GitHub Pages
+        id: deployment
```
henryiii, line 65:
```suggestion
        id: deployment
        if: >
          success()
          && github.ref == 'refs/heads/master'
          && github.repository == 'halogenica/beautifulhugo'
```

A workflow dispatch on a branch shouldn't trigger this by mistake.

henryiii:
Okay, this looks pretty good. You can change the actions settings then click squash & merge, and we can try it out.

Issue #455, opened 2023-01-22
Bigimg displays as empty div with Javascript disabled


WJBarnes456:
I noticed that, if Javascript is disabled, the bigimg on the home page of a website using this theme doesn't display, instead showing a blank box.

As an example, using the exampleSite with the triangle image commented out in config.toml and disabling Javascript using Noscript:

Javascript disabled (also note the empty description box on the right hand side):
![image](https://user-images.githubusercontent.com/39161512/213940035-dc1dcdba-d07f-4e1b-840b-a3516e7eed93.png)
Javascript enabled:
![image](https://user-images.githubusercontent.com/39161512/213940042-7cfb80b2-8efe-4970-89ad-7eacd93a59ce.png)

This is because the theme only populates a data div with the image data, not any elements on the page. The theme then relies on the ~Photoswipe~ (correction, main.js) JS dependency to populate the image, which can't run and has no fall-back in the case that Javascript is disabled in the user's browser.

WJBarnes456:
On the fork of this theme I use on my website, I've [applied a small patch](https://gitlab.com/WJBarnes456/beautifulhugo/-/commit/f1a5cdd0cd96c2ace49ad8b47c60c1d5c7d05b18) to fix this, though it isn't at full feature-parity with the JS version, as it only displays the first image, and doesn't iterate through them at all.

WJBarnes456:
I've also just noticed that the [base main.js code]( https://github.com/halogenica/beautifulhugo/blob/master/static/js/main.j) supports a couple of further features which my previous patch doesn't support: setting `background-position` when the position attribute is set, and building Markdown links from the src attribute (although I'm not convinced the [markdown parsing regex](https://github.com/halogenica/beautifulhugo/blob/master/static/js/main.js#L144) is correct, as it doesn't handle nested brackets)

WJBarnes456:
Closed by https://github.com/halogenica/beautifulhugo/commit/f0ff519ec72aaef0a4e4511c3e1099def1740d16

PR #456, opened 2023-01-22
Fallback to first bigimg when JavaScript disabled

WJBarnes456:
Solves https://github.com/halogenica/beautifulhugo/issues/455 by statically configuring the page source to use the source and description of the first image in bigimg. When the description is empty, also pre-hides the description box.

When JavaScript is enabled, this theme's main.js
(https://github.com/halogenica/beautifulhugo/blob/master/static/js/main.js) takes over and will overwrite this style attribute, changing the image in use.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -15,12 +15,12 @@
 
 {{ if or $bigimg $title }}
   {{ if $bigimg }}
-    <div id="header-big-imgs" data-num-img={{len $bigimg}} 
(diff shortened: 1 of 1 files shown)


WJBarnes456:
I'm happy with the no-JS version not including the link rendering, but there's a very simple extension to render _all_ Markdown in the desc attribute by piping `$firstimg.desc` into `markdownify`.

This would then introduce a feature disparity the other way around; I'm not sure if that's preferable or not, so I've omitted it from the PR.

henryiii (approved):

PR #464, opened 2023-04-25
Add support for figcaption tag

precisemath:
We are unable to see captions underneath images using the `` tag.
Hence added support for `figure` and `figcaption` tags. I tested this functionality on my website.


henryiii:
I think this works now, let me know if not!

PR #466, opened 2023-06-15
Update bootstrap to 5.3.0

### LINKED ISSUE #401: Plan to upgrade bootstrap?
The bootstrap used is v3.3, which is quite old, and it will be deprecated some time in the future, so any plan?

davidmn:
Aiming to resolve #401


henryiii (approved):
I tested this locally a bit and it seems fine.

PR #473, opened 2023-06-19
WIP: bootstrap 4

henryiii:
Starting a draft of bootstrap 4. The header is still broken, as is the removal of `pager` making the next/back buttons ugly.

Some help available at http://upgrade-bootstrap.bootply.com.

Followup to #466, #467, #471, #472.


henryiii:
narrow version needs bigimg fix

PR #481, opened 2023-11-10
Site.Author is deprecated, use Params.author instead

Rivierg:
When building I got a warning that `Site.Author` is deprecated.


henryiii:
It's annoying that users will need to move their author list, but not sure there's much that can be done to avoid that.

henryiii:
Thanks!

PR #486, opened 2023-11-30
Dev001

vinceyoumans:
I made these changes recently and seems to work with out giving me warnings.


henryiii:
Changes 180+ files and changes things like the name of the theme.

PR #489, opened 2024-02-11
Add Cusdis support

BarthPaleologue:
## What is Cusdis

Cusdis is an open source alternative to Disqus for website comment sections. It is lightweight and don't require the reader to create an account to leave comments.

## Implementation

Cusdis support in the theme is implemented in the same way as the existing support for Disqus.
 
The App ID must be specified in the `config.toml` at the root of the project


henryiii:
Thanks! Looking forward to trying it. Disqus is unusable these days, loaded in ads.

PR #495, opened 2024-04-23
Add about social icon

xuhdev:
This gives the website author some flexibility to put a link to biograph.


ibnunowshad:
Do you have some screenshot of this change for `fas fa-at`?

xuhdev:
Yes, check out a live example at https://hugo-apt.8hob.io/

ibnunowshad:
> Yes, check out a live example at https://hugo-apt.8hob.io/

It is more like a `@` for email. But I may use it for email instead of `About`

xuhdev:
I simply followed fontawesome's decision... I personally also got the impression of @ being "about". @ is not only used in email, but also indicating one's host generally (think about ssh, ftp). Today also used for pinging someone like how it is used on GitHub.

PR #496, opened 2024-05-07
seo/article: Fix dateModified and add timezone

jaller94:
* `.PublishDate` and `.Date` are the same on my blog. However, `.Lastmod` contains the last modified timestamp.
* https://search.google.com/test/rich-results/ said that it would like to see a timezone


henryiii (approved):

ibnunowshad:
How to verify this change?

PR #500, opened 2024-07-05
Added feature to disable dark mode

hydr0nium:
I added an additional css file `main-nodark.css` that has no darkmode feature inside to disable the automatic change to dark mode on some devices. This can be done through the `hugo.toml` file:

```toml
[Params]
  [...]
  disableDarkmode = true # Disable darkmode features / Force light mode on all devices.
  [...]
```

### DIFF
--- a/layouts/partials/head.html
+++ b/layouts/partials/head.html
@@ -64,7 +64,12 @@
   <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@3.4.1/dist/css/bootstrap.min.css" integrity="sha384-HSMxcRTRxnN+Bdg0JdbxYKrThecOKuH5zCYotlSAcp1+c8xmyTe9GYg1l9a69psu" crossorigin="anonymous">
   {{- end -}}
 
+
+  {{- if .Site.Params.disableDarkmode -}}
+	<link rel="stylesheet" href="{{ "css/main-nodark.css" | absURL }}" />
+  {{- else -}}
   <link rel="stylesheet" href="{{ "css/main.css" | absURL }}" />
+  {{- end -}}
 
   {{- if .Site.Params.staticman -}}
   <link rel="stylesheet" href="{{ "css/staticman.css" | absURL }}" />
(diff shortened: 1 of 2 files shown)


henryiii:
Is there a way to avoid doubling the number of CSS files we have to maintain? Also, what about only dark?

(Also, why not just not put your device in dark mode if you don't want dark mode... Guessing this is because a site might not have dark-ready images?)

hydr0nium:
@henryiii Yeah the doubling of the CSS files is unfortunate but as far as I know there is no way of templating in the static folder where the CSS files live. 

---

An option to turn the auto darkmode off would be the right decision in my opinion. Another idea would be a button on the site that switches between dark and light mode as there is no easy way in turning of the `@media` CSS rule for just one site. I like light mode on some sites and dark mode on others and believe everyone should have the choice to choose individually for each site instead of being force by their system settings. 

As for the `only dark mode` options I also found no proper way. It's even harder because the dark mode is baked into the media rules and I am not skilled enough with css to port everything over on my own. Plus I guess it would result in yet another CSS file. If anyone has an idea how we could fix the `multiple CSS files` issue I am all ears. 

Maybe the button that lets users switch is not such a bad idea, but again I am not skilled enough for that, because my expertise is not with CSS

dustymabe:
Can we get this merged?

datatrigger:
Hi @hydr0nium,

Thank you very much for your work, now my beautifulhugo websites renders correctly when the user's browser forces dark mode. On my side, I added support for code blocks and Chroma as well([PR](https://github.com/datatrigger/datatrigger/pull/2)), if anybody is interested.

Hi @henryiii, could this be merged?

henryiii:
The problem is a rather large new file to maintain, and why shouldn't there be a disable light mode version too (another file)?

(Also, I don't understand the need for every site to add their own toggle. I hate it when sites don't have an auto mode but do have a toggle - I don't want to have to flip it at night when I'm trying not to be flash-banged. I've even got one of those at https://scikit-build.org that I haven't been able to fix. :) If I don't want dark mode, I won't put my device in dark mode. But from an author's standpoint, it's a bit more work to make sure images and such look right in both modes.)

I want to see if there's a better way to do this without duplicating the file. But with AI it's pretty easy to make sure the files are in sync, so if I can't find a better way, I'd be okay with it.

PR #502, opened 2024-07-10
Require Hugo 0.120+, fix deprecation warnings

hydr0nium:
I changed the parameter .site.IsServer parameter to hugo.isServer because the first one is deprecated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,13 +87,21 @@ Client side highlighting does not require pygments to be installed. This will us
 
 ### Disqus support
 
-To use this feature, uncomment and fill out the `disqusShortname` parameter in `config.toml`.
+To use this feature add your disqus shortname to the hugo.toml file like this:
+
+```toml
+[services]
+  [services.disqus]
+    shortname = ''
+```
+
+For further reference see [hugo config](https://gohugo.io/methods/site/config/)
 
 ### Staticman support
 
-Add *Staticman* configuration section in `config.toml` or `config.yaml`
+Add *Staticman* configuration section in `hugo.toml` or `hugo.yaml`
 
-Sample `config.toml` configuration
+Sample `hugo.toml` configuration
 
 ```
 [Params.staticman]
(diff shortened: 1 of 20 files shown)


henryiii:
Any idea of how far back the new one works?

henryiii:
New in 0.120 https://gohugo.io/functions/hugo/isserver/

Did they have to deprecate the same version they added the replacement?

henryiii:
Eh, we are on 0.128 now, so I'd be okay with making 0.120 the minimum, I think. Anyone else?

hydr0nium:
@henryiii 

I would say that being basically 8 versions ahead of the deprecated function is enough.

---

I also found another deprecated feature .Site.DisqusShortname which also got deprecated since 0.120. If we change to 0.128 we can also update that as well.

henryiii:
Can we make it a hard error (errorf) if too old?

hydr0nium:
Could we use the min_version parameter in the [theme.toml](https://github.com/halogenica/beautifulhugo/blob/master/theme.toml) file?

layouts/partials/disqus.html
```diff
@@ -1,7 +1,7 @@
 {{ if (.Params.comments) | or (and (or (not (isset .Params "comments")) (eq .Params.comments nil)) (.Site.Params.comments)) }}
-  {{ if .Site.DisqusShortname }}
+  {{ if .Site.Config.Services.Disqus.Shortname }}
```
henryiii, line 2: Should we error if `Site.DisqusShortname` is set? Similar to what I added in #481? Otherwise no one will know this was changed, it will just vanish from sites.

hydr0nium: Yeah I guess you can add that. Sounds good.

hydr0nium:
> The theme.toml one sounds good. I assume Hugo would error with a nice message if that was too high.

I did some tests and changing `theme.toml` would result in telling the user that they are using the wrong version. The problem is though that it does not stop the build. 

Maybe adding another check in something like the `baseof.html` would be a second measure to stop the build.

layouts/partials/disqus.html
henryiii: If you'd like to, go ahead. Pretty distracted at the moment with catch up in various projects.

hydr0nium: My idea is to use this in the `baseof.html`
```html
{{ if le hugo.Version "0.120.0" }}
	{{ errorf "Please use Hugo version 0.120.0 or higher!"}}
{{ end }}

{{ if not (eq .Site.DisqusShortname "") }}
	{{ errorf "Your use of disqusShortname is deprecated by Hugo; Please reference the README.md for further instructions.
	This error message will be removed in the future because we are also using .Site.DisqusShortname for checking this."}}
{{ end }}
 ```

Comment (henryiii) [reply layouts/partials/disqus.html:2]: Seems reasonable. I can move the other check (Site.Author) there too after it's in.

**Comment by henryiii** [comment]
Would you mind rebasing or merging to resolve conflicts? Then I'll trigger the CI.

File: layouts/_default/baseof.html [hunk 2]
```diff
@@ -1,3 +1,12 @@
+{{ if le hugo.Version "0.120.0" }}
+	{{ errorf "Please use Hugo version 0.120.0 or higher!"}}
+{{ end }}
+
+{{ if not (eq .Site.DisqusShortname "") }}
+	{{ errorf "Your use of disqusShortname is deprecated by Hugo; Please reference the README.md for further instructions.
+	This error message will be removed in the future because we are also using .Site.DisqusShortname for checking this."}}
```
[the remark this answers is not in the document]
hydr0nium: I think the problem is somewhere in the single.html, but I can't find where.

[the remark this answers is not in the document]
hydr0nium: I will take a look soon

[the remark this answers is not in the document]
henryiii: It says line 6 of baseof.html, which is exactly this multiline string.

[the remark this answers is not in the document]
hydr0nium: Should be fixed now

README.md
```diff
@@ -87,13 +87,21 @@ Client side highlighting does not require pygments to be installed. This will us
 
 ### Disqus support
 
-To use this feature, uncomment and fill out the `disqusShortname` parameter in `config.toml`.
+To use this feature add your disqus shortname to the hugo.toml file like this:
+
+```toml
+[services]
+  [services.disqus]
+    shortname = ''
```
henryiii, line 95:
```suggestion
[services.disqus]
shortname = ''
```

Probably matches everything else, so not changing it here, but I really don't like this fake TOML nesting. Pretending TOML cares about indentation when it doesn't isn't helpful, IMO.

henryiii (approved):

henryiii:
Thanks for pushing this through!

PR #503, opened 2024-07-23
Update hugo.toml

fyordsvat:
Update config


hydr0nium:
Maybe we should combine this with #505 To make it consistent.

henryiii:
I think this is basically present now.

PR #504, opened 2024-07-23
Update theme.toml

fyordsvat:
correct hugo version


deining:
After this merge, when previewing the example site, get a warning:

```
WARN  Failed to read module config for "beautifulhugo" in "/home/me/beautifulhugo/theme.toml": "_stream.toml:7:20":
unmarshal failed: toml: float can have at most one decimal point
```

We should either put the version number in quotes or revert this merge IMHO.

henryiii:
Oops, you are right, this is TOML, not YAML. Versions are strings, so it should be a string, I believe.

henryiii:
Fixed

PR #509, opened 2024-07-31
Chore/upgrade deprecated versions

mikeyrcamp:


henryiii:
This makes tons of unrelated changes, including adding "forked" to the name!

PR #513, opened 2024-09-14
UI Ehanchement

androchentw:
* feat: use hreflang for SEO
* feat ui: Add custom embed multiple language newsletter
* fix ui: fix TOC
* fix ui: back-to-top due to no fadeOut and fadeIn in jquery.slim.min.js
* feat ui: Shorten the navbar after scrolling a little bit down


androchentw:
wrong

PR #515, opened 2024-11-01
Add Last.fm social icon

alexcannan:
I'm an avid Last.fm user. Noticed it was missing and font awesome conveniently already had the logo (in two variations!)

I also took the liberty of including it in `exampleSite/hugo.toml` so it's easily accessible to first-time users


henryiii:
Thanks!

PR #517, opened 2024-11-10
Update socials example TOML in README.md 

davidmn:
Changes the snippet of TOML to match how it works in [practice](https://github.com/halogenica/beautifulhugo/blob/fd74af48c3bb5e51187ca350324d2291de65f283/exampleSite/hugo.toml#L54C1-L54C16)

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ author:
 ```
 
 ```toml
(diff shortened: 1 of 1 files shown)


henryiii:
Thanks!

PR #520, opened 2024-11-15
Fix Bluesky icon and improve URL

davidmn:
My last PR to add Bluesky wasn't perfect. This change uses the correct icon and allows you to just set your handle rather than the full url.


TomCools (approved):

TomCools:
Was integrating this and noticed these issues, happy to already see a PR for this.

PR #521, opened 2024-11-21
Update font-awesome to 6.7.1

josephdecock:
Bluesky icons weren't working for me with the selfHosted option enabled. I downloaded the latest free version of font awesome and updated, and the bluesky icon worked. I'm not sure how these updates are normally handled and packaged (I notice that the copyright header is added in the diff, which makes me think maybe some additional processing is normally done).


josephdecock:
Another issue here is that there are many files in static/fontawesome/webfonts that I did not change. Those weren't in what I downloaded from FA, and I don't know where they come from or what they are for.

VincentTam:
thx 4 ur time & effort!
i'm trying to fix this cuz icons on my blog are gone
more updated FA version aside, [FA's docs](https://docs.fontawesome.com/web/setup/host-yourself/webfonts#custom-icons-in-downloaded-kits), for the Webfonts method doesn't recommend `all.css` in production 
https://github.com/halogenica/beautifulhugo/blob/ac686e94dffa0253dca90b9f31ee6b2d3b42e5fa/layouts/partials/head.html#L57-L65
I'm going to fix this

VincentTam:
i tried building change on top but i couldn't due to more recent changes in disqus template, causing error when i test `./exampleSite`
```
$ hugo serve -s "./exampleSite" --themesDir "../.." --logLevel info
Error: html/template:single.html:94:24: no such template "_internal/disqus.html"
vin100@vin100-81yq:~/beautifulhugo$ git grep disqus.html
layouts/_default/single.html:            {{ template "_internal/disqus.html" . }}
layouts/partials/disqus.html:      {{ template "_internal/disqus.html" . }}
```
i regret that i have to close this PR restart another one

henryiii:
Upgraded in #543, thanks for the help!

PR #528, opened 2025-02-20
fix for white header background with dark theme

sesom42:
fix for white header background with dark theme on devices with screen widths less than 365px


alexandear:
Hey, thanks for the fix! Could you please provide before-and-after screenshots of your changes?

sesom42:
These are screenshots of the theme demo website before and after the change on a device with 360px media width (Galaxy S20) and dark system theme.

Before (header unreadable):
grafik


After:
grafik

PR #532, opened 2025-04-12
Modify deprecated _internal namespace (since Hugo v0.146.0)

sutehachi:


aloisdg (approved):
LGTM

henryiii:
Does this work on older Hugo’s too? We try to keep compat with a few older versions.

jmooring:
The proposed change creates an infinite loop.

You need to rename the wrapper from `discus.html` to something like `disqus-wrapper.html`.

Then, in `disqus-wrapper.html` call the embedded partial:

```text
{{ partial "disqus.html" }}
```

Finally, in `single.html`, call the wrapper (line 94):

```text
{{ partial "disqus-wrapper.html" . }}
```

Issue #562, opened 2026-03-01
Summaries showing images twice


ViniciusMenezesDev:
I'm new to Hugo and I'm migrating a blog from Jekyll to Hugo. Many of my posts start with an image, which often is also the post's featured image. This is causing the index post list to show images twice:

```md
---
date: "2026-01-30T20:00:00Z"
image: /files/2026/01/wifi-analyzer.jpg
title: 'Como configurar a frequência do ponto de acesso Wi-Fi para reduzir a interferência e melhorar o sinal - parte 1: faixa de 2,4 GHz'
---

{{< image src="/files/2026/01/wifi-analyzer.jpg" >}}

Seu celular lista várias redes Wi-Fi onde você mora? Como saber se as redes dos vizinhos não estão interferindo na sua? Em grandes cidades com cada vez mais prédios, esse cenário é cada vez mais comum. Este artigo apresenta um aplicativo que te permite analisar a ocupação do espectro Wi-Fi e sugere a melhor configuração para o seu roteador.

<!--more-->

As tecnologias sem fio se tornaram onipresentes: celulares **[4G]**, _notebooks_ com **[Wi-Fi]**, **[TVs]** analógicas, digitais, fones de ouvido **[Bluetooth]**, rádios **[AM]/[FM]**. Certamente você usa no seu dia-a-dia senão todas, algumas dessas tecnologias. No passado, os aparelhos eram conectados principalmente por fios, que geravam bagunça e poluição visual. Hoje, a bagunça e a poluição continuam existindo, só não são mais visíveis: os aparelhos atuais se conectam principalmente por **[ondas de rádio]**.
```

<img width="1207" height="1290" alt="Image" src="https://github.com/user-attachments/assets/ecb27481-aabf-40d7-a731-36268680e747" />

henryiii:
Aren't you listing the image twice, though? Once in the front matter, and once in the summary? Couldn't you remove one of them? It seems like it would be better to find some way to control placing the image in the article (if that's the problem) rather than stripping out *everything*, like links (which are pretty common in summaries!).

Bas-Man:
I feel this is merely a misunderstanding of how frontmatter: "image" works. You should simply choose one or the other option.  As @henryiii points out. You are basically referencing the image twice. Once through the default layout system and again directly. Choose ONE.

PR #563, opened 2026-03-01
Summaries were showing images twice #562

ViniciusMenezesDev:


henryiii:
I don't think this is the right fix - you are listing images twice, so that's why they are being shown twice. I think we might instead need to see why you can't just list the image once.

PR #564, opened 2026-03-02
Feature: Set controls on pages to prevent or limit page overview summaries

### LINKED ISSUE #559: Meta tag and div support to block Search Engine AI summary results / snippets
I was wondering if there were any plans to add an AI search result summary control for blog posts and or pages.

Based on this article, it appears that we can potentially use meta tags and a div tag to restrict the size of AI snippets or prevent the page from having a snippet generated.

[See article here](https://www.playwire.com/blog/how-to-block-google-ai-overview-from-using-your-content)

I was considering adding the following

```toml
[googleAISearch]
MaxSnippet = default # -1, other values, 100, 150, 300
```

There can also be a frontmatter which sets an overrides the  ```.Site.Params.googleAISearch``` for the given page.

Also, a shortcut could be created to surround a section of the page to prevent that segment from being included in the summary.

Looking for some thoughts..

Bas-Man:
Adds a configuration option to *hugo.toml* 
Adds new partial which can be customised *meta-ai-search-summary-limits.html*
Adds a new shortcode *no-ai-summary*
Adds an explainer page as a new post.

Closes #559


Bas-Man:
Before moving forward with this. I would like some ideas on any better naming for this features in the Hugo.toml and the page.Params. I am not really happy with these names and I also know they are not really consistent.

Thoughts please.

Bas-Man:
Hi @henryiii, I think this is ready for a review. I would also appreciate any comments.

The pull request became a little more complicated as I added the ability to manage standard robot meta data.
My thought process was that it would be best to combine the typical robot meta data with the ai-summary early rather than having to tack it on later.

henryiii:
Looks pretty good, only one thought: Why does the index override Hugo.toml? Shouldn't it just affect the page it's on, like all the others? It's not supported for nested directories?

exampleSite/content/post/2026-02-27-meta-ai-search-summary-limits.md
```diff
@@ -0,0 +1,123 @@
+---
+title: Manage robot meta tags an limit AI Overview from using page contents
+date: 2026-04-14
```
henryiii:
```suggestion
date: 2026-05-10
```

(so it will show up at the top)

PR #571, opened 2026-03-13
Update Organziation.html partial

Bas-Man:
chore: formatting

Remove an empty lines that might appear if imformation is missing

fix: end of line

add: Only add organization information if organizationName exists

This avoids having a null name an unneeded ld-json entry in the site


Bas-Man:
I think completely removing the organisation if it is not defined might be a step too far.

Instead, I think I should use the `Params.Author.name` as organisation default if `orgranisationName` is not defined.

Looking for comments.

PR #587, opened 2026-05-05
fix SEO with built-in template

pad92:
fix og:type for article


henryiii:
I believe this was added in [v0.147.0](https://github.com/gohugoio/hugo/releases/tag/v0.147.0). We just bumped the minimum to 0.146, maybe we could go one higher.

pad92:
Hello, Yes, this has been added in [v0.146.0](https://github.com/gohugoio/hugo/blob/v0.146.0/tpl/tplimpl/embedded/templates/_partials/opengraph.html)

> Reimplement and simplify Hugo's template system https://github.com/gohugoio/hugo/commit/83cfdd78ca6469e6d7265323d9fad1448880e559 @bep

henryiii:
Ahh, perfect, thanks!

PR #624, opened 2026-05-08
feat: beautifulfigure

henryiii:
Adding a new shortcode, and a way to get normal figures.

- **feat: support beautifulfigure**
- **feat: setting to get standard figures too**

Assisted-by: OpenCode:Kimi-K2.6


layouts/partials/shortcodes/figure.html
henryiii, line 1: This doesn't need the outer check, will fix later.

PR #730, opened 2026-05-15
feat(layouts): allow custom home title in navbar

pad92:
Update the navigation bar to use `homeTitle` from site parameters if defined, falling back to the default site title. This provides more flexibility in controlling the brand name displayed in the header.

### DIFF
--- a/layouts/partials/footer.html
+++ b/layouts/partials/footer.html
@@ -69,11 +69,12 @@
             {{ .Site.Lastmod.Format "2006" }}
           {{ end }}
 
-          {{ if and .Site.Params.author.name .Site.Title }}
+          {{ if and .Site.Params.author.name .Site.Params.homeTitle | default .Site.Title }}
             &nbsp;&bull;&nbsp;
           {{ end }}
           {{ if .Site.Title }}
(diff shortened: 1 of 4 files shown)


layouts/partials/footer.html
```diff
@@ -69,11 +69,12 @@
             {{ .Site.Lastmod.Format "2006" }}
           {{ end }}
 
-          {{ if and .Site.Params.author.name .Site.Title }}
+          {{ if and .Site.Params.author.name .Site.Params.homeTitle | default .Site.Title }}
             &nbsp;&bull;&nbsp;
           {{ end }}
           {{ if .Site.Title }}
-            <a href="{{ "" | absLangURL }}">{{ .Site.Title }}</a>
+            <a href="{{ "" | absLangURL }}">{{ .Site.Params.homeTitle | default .Site.Title }}
```
henryiii: The condition on this statement is "if .SiteTitle", they mismatch now.

pad92: > The condition on this statement is "if .SiteTitle", they mismatch now.

yes, i see it.
do you prefer like that ?

henryiii:
Thanks!

PR #734, opened 2026-05-17
style(post-meta): add text-reset class to author links

pad92:
Add the `text-reset` CSS class to author links in the post metadata partial to ensure they inherit the surrounding text color and maintain visual consistency.


henryiii:
Thanks! I didn't even notice that.

PR #737, opened 2026-05-18
fix: Humanize recipe times

Bas-Man:
Humanize the recipe times.

- This just provides the intial ability to show recipe times in a human readable format.

- It might be worthwhile converting hours to minutes to prevent the recipe width from growing too much.


henryiii:
Thanks!

PR #739, opened 2026-05-18
fix: correct paths for subdirectory deployments

henryiii:
:robot: *Human guided, AI assisted PR ([using this skill](https://github.com/henryiii/skills/blob/main/branch-and-pr/SKILL.md)). AI text below.* :robot:

## Summary

Fixes path issues that break the site when deployed under a subdirectory (e.g. `halogenica.net/beautifulhugo/`):

- **`index.json` loaded from base URL**: `search-fuse.js` had a fallback `|| '/index.json'` that bypassed the `relURL`-computed path from the template, causing the search index to be fetched from the domain root instead of the subdirectory.
- **`false` in commit link URL**: When `commit = false` in config (and `enableGitInfo = true`), the footer rendered `false` + GitInfo hash as the link URL (e.g. `false87df0540...`). Now the commit link only renders when `.Site.Params.commit` is actually set to a truthy value.
- **Redundant `and` in show-source.html**: `{{ if and $.GitInfo }}` simplified to `{{ if $.GitInfo }}` (single-condition `and` is a no-op).

Assisted-by: opencode:glm-5

### DIFF
--- a/layouts/partials/search-provider.html
+++ b/layouts/partials/search-provider.html
@@ -1,4 +1,4 @@
-{{- $provider := "fuse" -}}
+{{- $provider := "none" -}}
 
 {{- with site.Params.search -}}
   {{- $provider = lower (.provider | default (.providor | default "fuse")) -}}
--- a/layouts/partials/search-provider.html
+++ b/layouts/partials/search-provider.html
@@ -1,4 +1,4 @@
-{{- $provider := "none" -}}
+{{- $provider := "fuse" -}}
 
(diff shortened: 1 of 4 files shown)


layouts/partials/search-provider.html
```diff
@@ -1,4 +1,4 @@
-{{- $provider := "none" -}}
+{{- $provider := "fuse" -}}
```
henryiii:
```suggestion
{{- $provider := "none" -}}
```

PR #747, opened 2026-05-20
feat(search): decode HTML entities in search results

pad92:
Add a `decodeHtml` helper function to ensure that HTML-encoded characters in page titles and excerpts are properly decoded before being escaped and rendered in the search results dropdown. This prevents double-encoding or raw entity display in the UI.

### DIFF
--- a/static/js/search-fuse.js
+++ b/static/js/search-fuse.js
@@ -163,9 +163,7 @@
 
   function decodeHtml(text) {
     if (!text) return '';
-    var div = document.createElement('div');
-    div.innerHTML = text;
-    return div.textContent || div.innerText;
+    return new DOMParser().parseFromString(text, 'text/html').body.textContent || '';
   }
 
   function escapeHtml(text) {
--- a/static/js/search-fuse.js
+++ b/static/js/search-fuse.js
@@ -161,6 +161,13 @@
     }
   }
 
+  function decodeHtml(text) {
+    if (!text) return '';
+    var div = document.createElement('div');
+    div.innerHTML = text;
+    return div.textContent || div.innerText;
+  }
+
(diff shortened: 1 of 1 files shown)


henryiii:
My only question (which I can check later) is if it makes more sense to process this in the template, before we store it, or we are supposed to store as HTML and decode it here in the JavaScript.

pad92:
Replace innerHTML-based decodeHtml with DOMParser for XSS safety

The previous implementation of decodeHtml used innerHTML to decode
HTML entities:

    var div = document.createElement('div');
    div.innerHTML = text;                    // ← DOM XSS risk
    return div.textContent || div.innerText;

Assigning untrusted or partially-trusted text to innerHTML causes the
browser to parse and execute any embedded HTML, including inline event
handlers (e.g. ``), before textContent can
extract the plain text. This is a classic DOM-based XSS vector.

The fix replaces it with DOMParser.parseFromString():

    return new DOMParser().parseFromString(text, 'text/html')
      .body.textContent || '';

DOMParser operates in a detached document context: scripts are not
executed and event handlers never fire, making it safe to parse
arbitrary HTML strings. It correctly decodes all named HTML entities
(including extended ones such as &rsquo;, &ldquo;, &ndash;) that
Go's html.UnescapeString does not handle, and that Hugo's plainify
leaves intact in the search index JSON.

The call chain for rendering a search result is therefore:
  JSON value (may contain &rsquo; etc.)
    → decodeHtml()   : safely decode all HTML entities via DOMParser
    → escapeHtml()   : re-encode as safe HTML via textContent/innerHTML
    → innerHTML      : insert the now-clean markup into the DOM

henryiii:
Thanks!

PR #750, opened 2026-05-29
fix: home and nodes incorrectly return the same canonical url

Bas-Man:
The previous code returns the same canonical url

- http://something/post/

This canonical reference remained the same whether you were really on "/" or "/page/2" or three.

This is an issue as search engines would see "/post/" and "/post/page/2" as duplicates, impacting page ranking.

The new code resolves these issues and correctly gives each node its correct canonical URL.

### DIFF
--- a/layouts/partials/seo/main.html
+++ b/layouts/partials/seo/main.html
@@ -1,4 +1,11 @@
 {{- partial "seo/schema" . }}
 {{- partial "opengraph.html" . }}
-<link rel="canonical" href="{{ .Permalink }}" />
 {{- partial "seo/twitter" . }}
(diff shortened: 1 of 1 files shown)


Bas-Man:
I have found an issue with my template for this code. The timing of pagination is different on my site, even though I am using the default templates. I need to work out why this happens to ensure that this fix works generically.

It works correctly with the example site. But in my case. My landing page changes, meaning I do not get the paginated /post/ page.

Based on what I have, I have to ensure that the pagination I am creating occurs after the one that takes place in "layouts/index.html"

Bas-Man:
Closing this as @henryiii is working on this with #753

Issue #1, opened 2016-03-16
Add this theme to the theme gallery


digitalcraftsman:
Hello @halogenica,

I'm sure you want to add your theme to Hugo's [theme gallery](http://themes.gohugo.io/). You can do this by adding this repository as submodule to the [hugoThemes](https://github.com/spf13/hugothemes) repository.

However, to make sure that the demo of you theme can be generated we need to add two things:

- within the `exampleSite` folder you need to have a directory named `static`, even if it's empty and not used. Since Git needs something to track inside that folder add an empty file like `.gitkeep` to this directory.
- we need some metadata about this theme. Those are defined in the `theme.toml`. You can find an example [here](https://github.com/spf13/hyde/blob/master/theme.toml)

halogenica:
Thank you! Opened an issue to request it to be added as a submodule here:
https://github.com/spf13/hugoThemes/issues/118

digitalcraftsman:
Great.

PR #756, opened 2026-06-07
Move to graph and fix breadcrumb with pagination

Bas-Man:
This pull request contains a number of important fixes for breadcrumbs, author and other attributes.

1. Breadcrumbs were actually being created incorrectly and not unique across paginations
    The breadcrumb @id was the same for "/" and /page/2 on the home page. This has been addressed
    The true homepage should not have any breadcrumb, as breadcrumbs tell search engines the page path          
    from the root page to the current page. The previous code was some what circular in nature. Breadcrumbs 
    SHOULD exist for /page/2 and /page/3 and so on.

2. If an article contained multiple authors, they were not assigned truely unique @ids. This has been
    addressed

3. Add CollectionPage to .IsNode pages. Collection pages do not require a WebPage object to be present as it provides the same function.


Bas-Man:
@henryiii This is ready for review. I am not sure why, but I cant seem to request reviews.

henryiii:
Starting with an automated review from Claude:

Confirmed: `seo/schema` runs before `seo/canonical`, so its `.Paginate` call wins. Here is my review.

---

# Code Review: PR #756 — Move to `@graph` and fix breadcrumb with pagination

## Overview
Consolidates the previously-separate JSON-LD `<script>` blocks (WebSite, Organization, WebPage, Breadcrumb, Article/Recipe) into a **single `@graph`** document, cross-referenced by `@id`. Also reworks breadcrumb generation to be pagination-aware and adds a new `CollectionPage` node for list pages. I built the example site (`hugo --minify -s exampleSite`, v0.162.1) and inspected the emitted JSON-LD across home, paginated home, section, paginated section, taxonomy, and article pages.

**The `@graph` direction is genuinely good and well-executed in places**, but there are three correctness bugs — one of which contradicts the PR's own stated goal, and one that's a latent regression of a documented design invariant.

## What's good ✅
- **Single `@graph` consolidation** is the right SEO model — entities reference each other by `@id` instead of duplicating objects. Verified the article graph emits `WebSite / Organization / BreadcrumbList / WebPage / Person / Article` as distinct top-level nodes.
- **Per-author `@id`** (`#author-<urlize name>`) genuinely fixes the bug where multiple authors all shared `#author`. 
- **`publisher` / `isPartOf` / `mainEntityOfPage` now reference shared `@id`s** instead of inlining a duplicate Organization+logo on every article. Good dedup.
- The `{{ return ($x = $x | append (slice $schema)) }}` trick to return *multiple* nodes from `article.html`/`recipe.html` **works** — Hugo's `append` flattens the slice, so Person and Article land as sibling nodes (verified, no nesting).

## Bugs 🔴

### 1. Pagination single-source-of-truth violation (latent regression)
`schema.html:31` and `collection-page.html` each call `.Paginate` with their *own* page collections, in `<head>`, **before** `canonical.html:25` and the body `list.html`, which both paginate via `func/list-pages.html`. Hugo caches one paginator per page — first call wins, later calls are silently ignored. This is documented verbatim in `func/list-pages.html:6-10`, which exists *specifically to prevent this*. Since `seo/main.html:1` runs `schema` before `canonical` (line 3), **the SEO partial now hijacks the page's real pagination.**

The collections differ in a way that matters:
- `list-pages.html`: `where … "Type" "in" mainSections` **and drops `Params.hidden == true`**
- `schema.html:24`: `where .Site.RegularPages "Section" "in" .Site.MainSections` — **no hidden filter**, and keys on `Section` not `Type`

So hidden pages can reappear in the home listing and the page counts/URLs can diverge from the canonical link. The example site masks this (its only hidden page isn't in a main section), but the invariant is broken. **Fix:** route `schema.html` and `collection-page.html` through `partial "func/list-pages.html" .` instead of re-implementing the collection.

### 2. Home `/page/2/`, `/page/3/` get no breadcrumb — contradicts the PR's stated goal
`schema.html:21` computes `$isAbsoluteHome` from `$paginator`, but `$paginator` is still `""` at that point — it isn't assigned until line 31. So `(not $paginator)` is always true and `$isAbsoluteHome == .IsHome`, which is `true` on paginated home pages too. **Verified:** `page/2/` and `page/3/` emit no `BreadcrumbList`. The PR description explicitly says "Breadcrumbs SHOULD exist for /page/2 and /page/3." **Fix:** move the `$isAbsoluteHome` assignment to *after* `$paginator = .Paginate …`.

### 3. Duplicate breadcrumb entry on paginated section/term pages
Verified on `post/page/2/`, positions 2 **and** 3 are both `"Posts" → /post/`. The ancestors loop already emits the section, then the deep-page block (`breadcrumb.html`, the `if not .IsHome` branch) re-appends `.Title`/`.Permalink` — and `.Permalink` resolves to `/post/`, not the paginated URL. Result is a malformed list with a repeated node. **Fix:** drop the redundant group-title append (the ancestor already covers it), keeping only the final "Page N" item.

## Cleanup 🟡
- **`old-breadcrumb.html` is dead code** — a verbatim backup of the old partial. Remove it before merge.
- **`collection-page.html` hardcodes `where … "Section" "post"`** (with a comment admitting the assumption). It's also effectively dead: by the time it runs the paginator is already initialized by `schema.html`, so its `.Paginate $posts` arg is ignored and returns the cached paginator. Misleading — use `func/list-pages.html`.
- **`schema.html:1` `$currentPageUrl := .Permalink`** is never used (the value actually used comes from `.Store.Get "currentPageUrl"`). Remove.
- Cross-partial state via `.Store.Set`/`.Store.Get` (`isAbsoluteHome`, `paginator`, `currentPageUrl`) is fragile coupling; passing a context dict would be clearer. Minor.

## Risk summary
- Bugs #2 and #3 are pure JSON-LD correctness (no runtime breakage) but produce wrong/duplicated structured data — and #2 fails the PR's headline goal.
- Bug #1 is the one to worry about: it silently overrides the deliberately-centralized pagination and can change the actual rendered listing. **This should block merge until schema/collection-page reuse `func/list-pages.html`.**
- No test coverage exists (theme convention); I'd recommend at minimum eyeballing `page/2/` breadcrumbs and a `hidden`-flagged main-section post after the fix.

Recommendation: **request changes** — fix #1 and #2 (and ideally #3) before merge; #1 by reusing `func/list-pages.html` resolves the paginate-ordering hazard wholesale.

henryiii:
(I can request Claude fix it, just let me know what you think of 1, 2, & 3).

Bas-Man:
Let me take a look at this as I disagree with some of its conclusions. I need to address the others that need adjustment.

Bas-Man:
I think I have addressed issues 2 and 3. 
How do you want to handle issue? I switched the code to use the list-pages partial. But I am not sure if it will limit itself to only post given that its mainSection might contain other sections.

Bas-Man:
I am not sure about the risks that co-pilot lists.

Bas-Man:
I think all the bugs have been addressed.

henryiii:
Yes, Claude is happy now. Asking it to do a bit of cleanup, then it's good to go.

Bas-Man:
@henryiii I have pulled these changes down and tested locally. Things are looking good. I have also started migrating my own schema to start using these from the current branch as an additional check and things are looking good.

henryiii:
Excellent, will merge then.

PR #758, opened 2026-06-08
fix: include a more marker in new post template

henryiii:
Explicit is better than auto, and this makes it easier.

### DIFF
--- a/archetypes/post.md
+++ b/archetypes/post.md
@@ -10,3 +10,5 @@ tags: []
 bigimg: []
 comments: true
 ---
+
+<!-- more -->


archetypes/post.md
```diff
@@ -10,3 +10,5 @@ tags: []
 bigimg: []
 comments: true
 ---
+
```
Bas-Man, line 13: I would include some text to simply state:

*Anything above this point will be included in post summary and will be visible on the listing page.*

henryiii: You'd have to delete that text every time, though. Everything else here is minimal (blank or generated to try to be correct).

Bas-Man: Fair call.

PR #791, opened 2026-09-02
refactor: move theme CSS/JS to assets/ and use Hugo Pipes

### LINKED ISSUE #790: Move static CSS/JS to assets/ and use Hugo Pipes

Split out from the last open item in #759.

The theme serves CSS and JS from `static/`, including pre-minified vendored copies. Moving the theme's own files (`main.css`, `codeblock.css`, `main.js`, and similar) to `assets/` and loading them with `resources.Get | minify | fingerprint` would give:

- cache busting through fingerprinted file names
- one source file per asset, with no hand-maintained minified variants
- SRI `integrity` attributes for free

Vendored third-party assets (Bootstrap, jQuery, Font Awesome, KaTeX, PhotoSwipe, Highlight.js, Mermaid) can stay in `static/` or move later. The `selfHosted` switch in `head.html` and `footer.html` must keep working in both modes.

The minimum Hugo version is 0.146.2, which supports all of this. This is a larger change and should be its own PR.

henryiii:

Closes #790.

Moves the theme's own CSS and JS (`main.css`, `dark.css`, `print.css`, `syntax*.css`, `codeblock.css`, `toc.css`, `search.css`, `staticman.css`, `hugo-easy-gallery.css`, `main.js`, `toc.js`, `search*.js`, `staticman.js`, `load-photoswipe.js`, `mermaid-render.js`) from `static/` to `assets/`.

New `css.html` and `js.html` partials load them with `resources.Get | minify | fingerprint` and emit `integrity` attributes. The partials accept a plain path or a dict with `id`, `media`, or `defer`.

Vendored third-party assets and the self-hosted fonts stay in `static/`, so `selfHosted` works unchanged in both modes. Users who override a theme stylesheet must now put the copy in their site's `assets/` directory instead of `static/`. Documented in `theming.md` and `CLAUDE.md`.

Verified the example site builds with `selfHosted` on and off and under a subpath `baseURL`.


henryiii:
Superseded by a PR from an upstream branch so a stacked PR can target it.

PR #794, opened 2026-09-02
feat: process page-bundle images with Hugo Pipes

henryiii:

Stacked on #793. Retarget to `hugo-pipes-assets` or `master` as the stack merges.

Images that Hugo can reach as resources (files in a page bundle, or under the site's `assets/` directory) are now processed at build time. Images under `static/` and remote URLs are served unchanged, so existing sites see no difference until they move an image into a bundle.

- New `func/image-resource.html` resolves a path to a processable image (page resource first, then `assets/`, skipping SVG). New `func/image-srcset.html` builds a `srcset` from `Params.imageWidths` (default 480, 800, 1200, 1600) without upscaling.
- **Post previews**: `thumbnail` / `image` resources are cropped to 192x192 and 384x384 with a 1x/2x `srcset`, `width`/`height`, and `loading="lazy"`.
- **bigimg**: page-resource header images wider than `Params.bigimgWidth` (default 1920) are resized down.
- **beautifulfigure**: resource images get a `srcset` and `sizes`, and `data-pswp-width`/`height` are filled from the image, so `size` is no longer needed.
- **gallery**: `dir` can now name a folder inside the page bundle. Square 600x600 thumbnails are generated, so no `-thumb` files are needed. The `static/` directory mode is unchanged.
- `Params.imageProcessing = false` turns all of this off.

The bundled example post gains a `thumbnail`, a resource figure, and a `photos/` gallery to demonstrate this. Docs added to `configuration.md`, `figures-and-galleries.md`, and `pages-and-layouts.md`. `exampleSite/resources/` (Hugo's image cache) is now gitignored.


henryiii:

Addressed both review points in `gallery.html`:

- Bundle galleries now collect resources regardless of `imageProcessing`. With processing off, the originals are served as thumbnails and links.
- Each image is checked before `.Fill`: SVGs (and everything when processing is off) are served unchanged and skip the PhotoSwipe size attributes, so a build no longer fails on an SVG in the folder.

Verified with a temporary SVG in the example gallery and with `imageProcessing = false`.

PR #821, opened 2026-09-15
fix(toc): render the panel from baseof so custom layouts get it

### LINKED ISSUE #818: Problem with table of content
I try to build a Hugo page using this theme and I have an example page that you can see the effect on.

# Problem description

You have the option to define a table of contents per page. Witht e default settings, in the top bar there is a TOC button that should open an overlayed box with the TOC. However, there seems to be an issue with this as the button has no effect on many pages.

# Example project

The source code is under https://github.com/stormshot-k10229/stormshot-k10229.github.io.git. ([Permalink](https://github.com/stormshot-k10229/stormshot-k10229.github.io/tree/69acdf58811d70329636ce18cfa5695ac3a44726) to current state)

## Steps to work

1. Clone the repo
2. Install a local hugo installation using `npm ci`
3. Install the hugo modules by `npm run hugo mod get`
4. Run the server using `npm run serve`

## Explain the problem

On the home site (`/`) the TOC button is non-functional.

If you go into the agreements page, it might work (or not). On https://stormshot-k10229.github.io/en/agreements/ it is working at the moment while on my freshly cloned project, it does not.

Within Agreements, the _Horses_ page also seems non-functional.

# Additional information

I can for sure deactivate the toc feature entirely. I will do so, probably. Just to have it working online.

The translation needs additional work but this is not needed to see the problem.

I am aware that the theme is not the latest version (suffixed by `/v5`) of the beautifulhugo theme. I tried updating but this did not bring any solution to the issue.

If the problem is caused by the page content, could you please explain what I need to do in order to get it running smoothly?

henryiii:

Fixes #818.

Sites that override `single.html` or `index.html` without calling the `toc.html` partial got a navbar toggle with no panel, so the button did nothing. The reporter's site does this on the home page and on single pages, which matches the pages they listed.

- Move `partial "toc.html"` from `single.html` and `list.html` into `baseof.html`, next to the nav that renders the toggle. The offcanvas is fixed-position, so its place in the DOM does not matter.
- Remove the orphaned toggle in `toc.js` if a page has no panel.
- Note the behavior in the configuration docs.

Verified by building the reporter's site against this branch: all three pages now render both the toggle and the panel.


christianlupus:
Thanks a lot!

PR #841, opened 2026-09-18
feat: add a video shortcode

### LINKED ISSUE #577: How can I add video content to my markdown pages
Is there a layout-keeping way to add video content to markdown files?

Using HTML-video tag breaks the layout and I wondered if there's some sort of plugin or different part of the hugo ecosystem that enables embedding video content.

Currently I added 

```html
<center>
<video controls autoplay="" preload="auto" width="320" height="240" poster="video/poster.png">
  <source src="video/video.mp4" type="video/mp4">
  Your browser does not support the video tag.
</video>
</center>
```
but would want to have some sort of meta configuration element that fits properly into the page layout.

Is there something already available?

Thanks

henryiii:

Adds a `video` shortcode for self-hosted video files. It renders a responsive `<figure>` with the browser's native player, an optional poster, caption, and attribution. Paths resolve like `beautifulfigure` (page bundle, `assets/`, `static/`, or remote URL) through a new `func/media-url.html` partial. `playsinline` is on by default and the figure is excluded from the PhotoSwipe scan. The fallback download link text is a new `videoUnsupported` i18n key in all languages.

The example site gets a `## video` section on the Shortcodes page with a live example, using a silent 38 KB ffmpeg-generated gradient clip.

Closes #303, closes #577


ottlinger:
@henryiii Autoplay is not good for SEO, therefore I did not activate it by default .....

henryiii:
It's off by default here, right?

ottlinger:
> It's off by default here, right?

Yes, sry - I misread the text, but the code says `false` by default. Thanks.

